PR #37, opened 2015-12-10
Allow to solve for vector valued matrix products

cptroot:
This pull request is inspired by wanting to solve matrix equations that are like
A x = b
When x and b are vectors of points in R^n. This is a first pass to get basic functionality, but there might be more places in the library where vector spaces could be used like this. Do you have anywhere else you think this could be useful?

### DIFF
--- a/src/sparse/linalg/cholesky.rs
+++ b/src/sparse/linalg/cholesky.rs
@@ -4,6 +4,7 @@ use std::ops::Deref;
 use std::ops::IndexMut;
 
 use num::traits::Num;
+use std::ops::{Add, Sub, Mul, Div};
 
 use sparse::csmat::{self, CsMat, CsMatView};
 use sparse::symmetric::is_symmetric;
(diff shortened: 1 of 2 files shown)


vbarrielle:
Hello, thanks for the PR.

While I agree that solving multiple right hand sides is a nice-to-have feature, there are a few issues with your current implementation:
* it's unlikely that a vector of right hand sides will be actually indexable to yield the values of similar rows.
* the copy bound on M is probably never achievable.
* even if the above points were to work, this would result in an API with lots of implicit copying

I think it's better to follow Rust's tendance to prefer explicit over implicit, and implement the multiple rhs feature as a dedicate `fn solve_multiple()` method, probably with a `IntoIterator<Item=V>` bound, that allocates the result once when starting the function, and the calls solve repeatedly on views of the input and the result.

vbarrielle:
I'm closing this PR to open it as issue #88 as it's a more general issue that needs dedicated abstraction to be properly handled.

Thanks for the input!

Issue #42, opened 2016-01-11
SpVecView


dylanede:
There's `SpMatView` for abstracting over matrix implementations, but no `SpVecView` for abstracting over vector implementations. When writing a function that takes vectors and/or matrices, the signature gets very long without using traits like this, since extra type parameters must be introduced specifying the storage details for each vector/matrix argument. `SpMatView` goes part of the way towards dealing with this, but `SpVecView` is really required as well.

vbarrielle:
Indeed, currently abstracting over sparse vectors is quite hard (but sparse vectors need more work in other aspects too, right now they're at the bare minimum to make the rest of the crate work).

Just out of curiosity, what kind of function are you trying to write that abstracts over both vectors and matrices?

Anyway, this is clearly something sprs should have in the future. Thanks for the feedback!

dylanede:
I'm looking at implementing some algorithms for constrained optimisation. For a quadratic program, for example, you need to be able to input _Q_, a sparse matrix, and _c_, a sparse vector, for the expression _x<sup>T</sup>Qx + c<sup>T</sup>x_. And you also need to input _A_ and _b_ for _Ax = b_, the equality constraint equation.

BTW, this is a great project. Rust really needs some decent scicomp libraries. Though it does seem that some things in the language are limiting what can be done right now in terms of ergonomics. Once abstract return types and impl specialisation land, some really good APIs should be possible.

vbarrielle:
Closed by #45

PR #45, opened 2016-01-11
add a trait to abstract over CsVec, see #42

vbarrielle:
@dylanede, would this be ok for your abstraction needs?


dylanede:
That looks great. Thanks.

vbarrielle:
Closing #42 then.

PR #51, opened 2016-01-12
Enhance documentation for the linalg module

### LINKED ISSUE #43: Documentation for algorithms in the `linalg` module
It would be helpful to have (even rudimentary) documentation for these functions, instead of having to figure out the correct usage from the test cases.

vbarrielle:
Fixes #43.

@dylanede is it ok with you?


dylanede:
The docs for the module and `LdlNumeric` look OK, but I'm not so sure about those for `LdlSymbolic`. When should someone use `LdlSymbolic` first instead of constructing an `LdlNumeric` directly? How is `LdlSymbolic` meant to be used? Its constructor takes a matrix, but to turn it into an `LdlNumeric` you give it yet another matrix. Is the `dim` method really the "number of dimensions" of a matrix? A matrix has two dimensions. Do you mean the rank of the matrix?

vbarrielle:
> When should someone use LdlSymbolic first instead of constructing an LdlNumeric directly? How is LdlSymbolic meant to be used? Its constructor takes a matrix, but to turn it into an LdlNumeric you give it yet another matrix.

Sometimes one might know the structure of the matrix in advance, but not its data. So the symbolic decomposition can be precomputed. This could not be done using `LdlNumeric`, because it might fail if the provided matrix is not SPD. `LdlSymbolic` does not have this issue though, as it won't look at the data.

> Is the dim method really the "number of dimensions" of a matrix? A matrix has two dimensions. Do you mean the rank of the matrix?

It is the number of dimensions, the matrix has to be square for the decomposition to make sense. But that should be more explicit.

dylanede:
I see, that makes it much more clear what is meant by "symbolic" in this case. Thanks.

Maybe I'm missing something, perhaps some terminology clash, but I still don't quite get what is meant by `dim`. A square matrix still has two dimensions, just they happen to have the same size. A vector has one dimension. A matrix has two.

vbarrielle:
You're right, I wanted to express in the API the fact that both dimensions were equal.

Maybe I could make it cleare by specifying this is the dimension of the right hand side the system can solve ? :)

dylanede:
So `dim` represents the length of a valid RHS vector? The length of `b` in `Ax = b`? `A` surely has to be square and symmetric for the construction of the `LdlSymbolic` to work, so why would the squareness be a property exposed by the API?

vbarrielle:
Because I thought a `fn shape(&self) -> (usize, usize)` would be even more confusing, as it would let the user wonder if the tuple values may become different.

dylanede:
What about `fn problem_size(&self) -> usize`? I see that you mean the number of "dimensions" of the space occupied by the vector with the name `dim`, but to me dimension is a rather overloaded term. Feel free to get more opinions about this though.

vbarrielle:
`problem_size` is a nice term, though a bit long, but I think I can live with that.

vbarrielle:
Updated. Feel free to provide additional feedback. Otherwise I'll merge it tomorrow morning.

PR #85, opened 2016-06-01
Error policy

vbarrielle:
Remove error codes that corresponded to programming errors. Only error cases left are:

- unsorted indices, since it will be possible to use this error in the future to create a sparse matrix with unsorted indices, which is functional though less performant.
- singular matrices in linear algebra, since this leaves the caller the opportunity to add some regularization.

The errors removed should be plain bugs when they happen (eg dimension mismatch, out of bounds, ...).

### DIFF
--- a/Guidelines.rst
+++ b/Guidelines.rst
@@ -1,0 +1,25 @@
+==========
+Guidelines
+==========
+
+This document describes development guidelines for this library.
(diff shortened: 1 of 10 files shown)


vbarrielle:
Fixes #85.

Issue #21, opened 2015-09-11
Bind to existing solvers


vbarrielle:
Offer optional bindings to SuiteSparse's solvers.

PR #103, opened 2017-01-04
Generic index types (WIP)

vbarrielle:
Leave the possiblity to specify the index type used by sparse matrices. This is a requirement for #21 and should also allow better memory usage when the dimensions of the matrix fit in a type thinner than a `usize`.

Checklist:

- [x] implement generic index for `CsMat`
- [x] implement generic index for `TripletMat`
- [x] implement generic index for permutations
- [x] find proper naming scheme for type aliases enabling to select the index type


vbarrielle:
Current status looks ok, remaining work will be planned in additional issues.

Issue #118, opened 2017-10-25
Implement alga traits


aepsil0n:
As motivated [in this thread](https://users.rust-lang.org/t/prior-work-on-krylov-subspace-methods/13261) on Krylov subspace solvers, I think it may be a good idea to try and see, whether this library's types could implement the traits provided by the [alga](https://github.com/sebcrozet/alga/tree/master/alga) library.

This would, of course, require some work. I would be happy to do that experiment myself, if you think it makes sense – at least as an optional feature.

vbarrielle:
Hello,

As I mentionned in the users.rust-lang.org thread, I'm not sure the traits provided by alga exactly match the requirements for Krylov subspace methods, but so far I've only looked at the matrix trait, so I could be very wrong, and you're very welcome to show me where I'm incorrect ;)

Still, implementing some interoperability trait seems like a good idea on its own, especially if it's done under an optional feature.

You might encounter some difficulties in the implementation though, from a quick glance scalars in alga require only `Clone`, but there are some spots in sprs where I relied on `Copy` for convenience. But changing that seems like a nice move too.

aepsil0n:
Hey, there! Thank you for the quick feedback :)

First of all, I completely agree that alga is not an exact fit of what we need for that specific problem. However, I do think that a proliferation of interoperability traits for each specific algorithm (or class of algorithms) may be a larger problem maintenance-wise than this slight mismatch. Thus, I would rather have a set of general algebraic abstractions. Alga does provide those to a certain extent. And I suppose it can be refined further, if that turns out to be necessary. After all, this use case perfectly fits its design goal, as far as I can tell.

For what I am doing practically, the type signature for my BiCGSTAB algorithm looks like this:

```rust
pub fn bicgstab<V, F>(guess: V, vector: V, matrix: F, tolerance: V::Field) -> V
    where V: FiniteDimVectorSpace + NormedSpace + Clone,
          V::Field: Identity<Multiplicative> + PartialOrd + Clone,
          F: Fn(&V) -> V,
```

I actually ended up not using the `Matrix` trait at all, as it seemed more complicated than what I needed. All the algorithm needs is the transformation associated with the matrix. Thus, I decided to pass that in as a `Fn(&V) -> V` instead of a matrix.

Nonetheless, the vector space and field abstractions are extremely useful. They reduce the amount of generics boilerplate, as they contain all the vector space operations. So from my perspective, implementing those vector abstractions takes priority.

[...]

PR #119, opened 2017-10-27
Add vectors by value

aepsil0n:
This is the first tiny step towards #118.

[alga's `ClosedAdd` trait](https://docs.rs/alga/0.5.2/alga/general/trait.ClosedAdd.html) requires addition by value. These implementations merely delegate to the by-reference implementation. I imagine that there is some space for optimization here later on, if you take ownership of the operands.

The same considerations apply for the other operators, too. I'd be happy to apply the procedure to all of them (could be done in this PR), but I would like to get feedback on this approach first.

### DIFF
--- a/src/sparse/vec.rs
+++ b/src/sparse/vec.rs
@@ -834,6 +834,51 @@ where N: Copy + Num + Default,
     }
 }
 
+impl<N, IS1, DS1, IS2, DS2> Add<CsVecBase<IS2, DS2>>
+for CsVecBase<IS1, DS1>
+where N: Copy + Num,
+      IS1: Deref<Target=[usize]>,
+      DS1: Deref<Target=[N]>,
+      IS2: Deref<Target=[usize]>,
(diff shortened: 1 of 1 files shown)


src/sparse/vec.rs
```diff
@@ -834,6 +834,51 @@ where N: Copy + Num + Default,
     }
 }
 
+impl<N, IS1, DS1, IS2, DS2> Add<CsVecBase<IS2, DS2>>
+for CsVecBase<IS1, DS1>
+where N: Copy + Num,
+      IS1: Deref<Target=[usize]>,
```
vbarrielle: `IS1` could be deref on a generic Index type, as in https://github.com/vbarrielle/sprs/blob/master/src/sparse/vec.rs#L814-L823

aepsil0n: Will fix it with the PR, while I am at it ;)

vbarrielle:
That's perfect, thanks a lot!

PR #120, opened 2017-10-27
Introduce a zero-sized vector as a "universal zero"

aepsil0n:
See the discussion in #118 for motivation.

~~`Zero` from num-traits has not been implemented yet~~

Also left out was the multiplication of that vector to a matrix from the left. It is more involved, as the left vector is treated like a matrix. A possible solution would be to extend the universal zero concept to matrices.

For now, the above step seems out-of-scope for this PR.

**EDIT:** changed my mind to include the `Zero` trait impl.


vbarrielle:
Looks all good, thanks!

PR #122, opened 2017-10-27
Alga integration: additive group for sparse vectors

aepsil0n:
So far for now. I added alga as an optional dependency that is part of the default features. The implementation so far only covers the algebraic structure up to an additive group. It was necessary to implement the negation operator for vectors as well.

Note that this is based on #120.

### DIFF
--- a/src/sparse/vec.rs
+++ b/src/sparse/vec.rs
@@ -993,6 +993,8 @@ mod alga_impls {
     }
 
     impl<N: Copy + Num, I: SpIndex> AbstractSemigroup<Additive> for CsVecI<N, I> {}
+
+    impl<N: Copy + Num, I: SpIndex> AbstractMonoid<Additive> for CsVecI<N, I> {}
 }
 
 #[cfg(test)]
--- a/src/sparse/vec.rs
+++ b/src/sparse/vec.rs
@@ -1012,7 +1012,11 @@ mod alga_impls {
               I: SpIndex
     {
         fn inverse(&self) -> CsVecI<N, I> {
-            -self.clone()
+            CsVecBase {
+                data: self.data.iter().map(|x| -*x).collect(),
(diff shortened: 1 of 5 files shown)


src/sparse/vec.rs
```diff
@@ -961,11 +976,84 @@ where IS: Deref<Target=[usize]>,
...
+    impl<N: Copy + Num, I: SpIndex> AbstractSemigroup<Additive> for CsVecI<N, I> {}
+
+    impl<N: Copy + Num, I: SpIndex> AbstractMonoid<Additive> for CsVecI<N, I> {}
+
+    impl<N, I> Inverse<Additive> for CsVecI<N, I>
+        where N: Clone + Neg<Output=N> + Copy + Num,
+              I: SpIndex
+    {
+        fn inverse(&self) -> CsVecI<N, I> {
+            -self.clone()
```
vbarrielle: Am I right in assuming this will create a zero element and then subtract `self`from it? I'm wondering if it wouldn't be better to explicitly return a new vector like this:

```rust
CsVec {
    data: self.data.iter().map(|x| -x).collect(),
    indices: self.indices.clone(),
    dim: self.dim
}
```

But it's possible that your version compiles down to the same code.

aepsil0n: No, this delegates to the `Neg` implementation, which modifies the values in-place, as it takes the vector by value. So all `inverse` does is to clone the vector and to negate its `data` attribute element-wise in-place. There is a copying vs. cache locality trade-off, but we are not going to save an allocation that way.

So… I did some benchmarks, your suggestion is indeed about 20% faster for a vector with about 10k entries. I am adding the change in a commit including the benchmarks.

aepsil0n:
Rebase & optimization are done.

vbarrielle:
Perfect, thank you for adding the benchmark!

PR #124, opened 2017-12-08
Fix to_dense for non square matrices

usamec:
Fixes small typo during dimension checking.


vbarrielle:
Good catch, thank you!

PR #126, opened 2018-01-17
Improve performance of sparse-dense operations

### LINKED ISSUE #125: Sparse vector dense vector product is painfully slow
Given:
```
let w = Array::range(0., 10., 0.00001);
let x = CsVec::new(1000000, vec![0, 200000, 800000], vec![1., 2., 3.]);
```

It is much faster to do:
```
x.iter().map(|(index, num)| num * w[index]).sum();
```
than
```
x.dot(&w);
```

This should either be fixed, or at least documented somewhere.

vbarrielle:
Fixes #125. See the issue for details of the investigation. @usamec I hope this addresses your pain points.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,3 +31,7 @@ tempdir = "0.3.5"
 [[bench]]
 name = "suite"
 harness = false
+
+[[bench]]
+name = "sparse_dense_products"
+harness = false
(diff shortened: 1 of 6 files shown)


usamec:
Let me check over weekend, but from the code it is looks good!

vbarrielle:
I've juste merged this and included it in version 0.6.1 on crates.io.

PR #131, opened 2018-08-27
change ndarray version from 0.10.0 to 0.11.2

hegza:
Using the old ndarray dependency conflicts with the most recent version.

Upgrade seems straightforward. All tests pass.


vbarrielle:
Hello, and thanks for looking into this.

I should merge this soon, but first I'd like to check if this requires bumping the version number.

vbarrielle:
After looking at ndarray's changelog, this should not be a breaking change. Thanks again!

PR #135, opened 2018-11-11
Improve verbosity of panic messages in CsMat

rth:
This improves verbosity of some panic messages in `CsMat`.

Currently, on the following example,
```rust
use sprs::CsMat;

let indptr = vec![0, 5, 8];
let indices = vec![761698, 328290, 828689, 761698, 780185, 901149, 780185, 144749, 258307];
let data = vec![1, 1, 1, 1, 1, 1, 1, 1, 1];
let a = CsMat::new_csc((2, 1000000), indptr, indices, data);
```

I'm getting an obscure `Indptr length does not match dimension` panic message, so this improves the verbosity of such messages a bit, to show the lengths.

### DIFF
--- a/src/sparse/csmat.rs
+++ b/src/sparse/csmat.rs
@@ -1077,14 +1077,26 @@ where
         let outer = self.outer_dims();
 
         if self.indptr.len() != outer + 1 {
-            panic!("Indptr length does not match dimension");
+            panic!(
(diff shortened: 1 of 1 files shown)


rth:
rustfm is failing on the code I have not changed (and I did run `cargo fmt` prior to commiting).

PR #137, opened 2018-11-11
run `cargo fmt` with Rust 1.30 (update from 1.29)

toastronics:
Fixes `cargo fmt --check` error encountered in PR #135


rth:
Thanks @toastronics !

vbarrielle:
Thanks for the fix. It looks like rustmft changes quite a bit between rust versions, I'm testing it on rust beta and the formatting changes again... Maybe we should select a particular rust version and only run rustfmt on that version. Stable seems to be too much of a moving target.

Issue #138, opened 2018-12-20
Support (de)serialization using serde-rs/serde


tomtung:
Currently serializing structs that own sprs sparse matrices seems quite annoying. Support for (de)serialization using https://github.com/serde-rs/serde would be really handy, and shouldn't be difficult.

PR #139, opened 2018-12-22
Add serde (de)serialization support

tomtung:
This addresses #138.


tomtung:
I think travis checks failed because the behavior of `cargo fmt` has changed in Rust 2018, which would modify old code if it's run.

vbarrielle:
I'm not familiar with serde yet, but I'm wondering: can one save a `CsMatView` and then deserialize it into a `CsMat`? In other words, what kind of type information leaks into the serialized representation?

Also, could you add tests so that I can see examples of using this serialization? Thanks!

PR #140, opened 2018-12-23
Support recycling internal containers of vectors & matrices

tomtung:
The obvious use case is when one only wants to keep some internal storage of a vector / matrix object (e.g. non-zero values stored in `data`) and doesn't need the object itself any more.

This also makes vector / matrix transformation more flexible. E.g. one might wants to map inner indices with some bijective function, and the resulting indices could go beyond the original dimension, which would cause `modify` to panic. With `into_raw_storage`, one can pull out the guts of a matrix / vector object, perform arbitrary transformations while avoiding memory copying / allocation if possible, and create a new object using these old storage objects when they are done.

### DIFF
--- a/src/sparse/csmat.rs
+++ b/src/sparse/csmat.rs
@@ -816,6 +816,27 @@ where
         &self.data[..]
     }
 
+    /// Destruct the matrix object and recycle its storage containers.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// use sprs::{CsMat};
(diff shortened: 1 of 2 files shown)


tomtung:
I think travis checks failed because the behavior of `cargo fmt` has changed in Rust 2018, which would modify old code if it's run.

vbarrielle:
Hello and thanks for this pull request. This is a nice addtion. Could you rebase onto the latest master that includes the rustfmt fixes from #141?

vbarrielle:
Perfect, thanks.

PR #141, opened 2018-12-23
update to rustfmt 1.0 (Rust 1.31)

toastronics:
[Rustfmt 1.0](https://blog.rust-lang.org/2018/12/17/Rust-2018-dev-tools.html#rustfmt)

Now that Rustfmt is 1.0, this should be the last regular update required to keep up with Rust versions for a while (update may be required for the eventual Rustfmt 2.0 release expected "some time in late 2019"). Rustmft will continue to add new features, but they will be featured gated for all Rustfmt 1.x versions.

### DIFF
--- a/.travis_rustfmt
+++ b/.travis_rustfmt
@@ -5,6 +5,6 @@ set -ev
 if [ -z ${RUN_RUSTFMT+x} ]; then
     echo "Not on stable channel: skipping rustfmt."
 else
-    rustup component add rustfmt-preview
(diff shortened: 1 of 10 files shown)


vbarrielle:
This looks good, thanks. I hope that's indeed the last change in rustfmt. It looked like the stability guarantees for rustmft 1.0 would not apply with a non-empty rustfmt.toml, but having only `max_width = 80` in it may be ok.

PR #143, opened 2019-01-04
Make construction functions more generic

tomtung:
Would have done the same for `csr_from_dense` and `csc_from_dense` if they were inside the `impl<N, I: SpIndex> CsMatBase<N, I, Vec<I>, Vec<I>, Vec<N>>` block.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -112,9 +112,7 @@ pub mod vec {
     };
 }
 
-pub use sparse::construct::{
-    bmat, csc_from_dense, csr_from_dense, hstack, vstack,
-};
+pub use sparse::construct::{bmat, hstack, vstack};
 
 pub use sparse::to_dense::assign_to_dense;
 
--- a/src/sparse/construct.rs
+++ b/src/sparse/construct.rs
@@ -1,13 +1,9 @@
 //! High level construction of sparse matrices by stacking, by block, ...
 
 use indexing::SpIndex;
-use ndarray::ArrayView;
-use num_traits::{Num, Signed};
-use sparse::csmat::CompressedStorage;
 use sparse::prelude::*;
 use std::cmp;
 use std::default::Default;
-use Ix2;
 
 /// Stack the given matrices into a new one, using the most efficient stacking
 /// direction (ie vertical stack for CSR matrices, horizontal stack for CSC)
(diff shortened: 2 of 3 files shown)


vbarrielle:
Thanks for this pull request.

I agree that it's a good idea to make the `vstack` and other similar constructors to be generic over the index type. I had forgotten about these constructors when I introduced the genericity over the index type.

However, for constructors such as `eye` and `zero`, I'd rather not have one be required to explicitly specify the index type if does not care. Unfortunately there is no way to have a default type parameter on a function definition or impl block. That's why when I introduced the generic index type I let these functions create matrices with the usize type. That's a bit unfortunate as it means it's a bit more cumbersome and less efficient to get another index type as one has to call `to_other_types`. One alternative could be to add another, more generic constructor (which would avoid a breaking change). Do you have any other suggestion?

tomtung:
Hi, I don't think this is a problem, because the type parameters are actually on the class `CsMatBase`, not on the methods `eye` and `zero`. Therefore it's still very convenient to create matrices with index type of `usize` by using `CsMat`:
```rust
let m: CsMat<f32> = CsMat::eye(3);
```
and only use `CsMatI` to ask for matrices with different index types:
```rust
let m: CsMatI<f32, u16> = CsMatI::eye(3);
```

The problem with the current behvaior is that `CsMatI::<f32, u16>::eye(3)` actually has the return type `CsMatI<f32, usize>`, which is very counter-intuitive:
```
error[E0308]: mismatched types
 --> src\main.rs:4:31
  |
4 |     let m: CsMatI<f32, u16> = CsMatI::<f32, u16>::eye(3);
  |                               ^^^^^^^^^^^^^^^^^^^^^^^^^^ expected u16, found usize
  |
  = note: expected type `sprs::sparse::CsMatBase<_, u16, std::vec::Vec<u16>, std::vec::Vec<u16>, _>`
             found type `sprs::sparse::CsMatBase<_, usize, std::vec::Vec<usize>, std::vec::Vec<usize>, _>`
```

For the same reason, I think it'd be so much more convenient if `csr_from_dense` and `csc_from_dense` were also methods inside the impl block:
```rust
impl<N, I: SpIndex> CsMatBase<N, I, Vec<I>, Vec<I>, Vec<N>>{
// ...
}
```

vbarrielle:
You're right, that is not a problem, and the current situation is rather problematic. If you want to, you can migrate the `csr_from_dense` and `csc_from_dense` in the appropriate impl block. Otherwise I'll merge as-is and do it myself later on.

Thanks again!

tomtung:
Pushed :)

vbarrielle:
Perfect, thanks!

PR #144, opened 2019-01-05
expand ndarray version req: work with 0.11.x and 0.12.x. Fixes #142

### LINKED ISSUE #142: Generalize ndarray version support
I've manually run `cargo test` with all ndarray versions that fall in the range:
`ndarray = ">=0.11, <0.13.0"` -- they all compile and pass test.  Could we generalize the ndarray version support to cover this?

A few people I work with have updated ndarray and gotten 'trait not implemented' errors because the top-level ndarray version doesn't match the sprs version.

(edited based on @rth's suggestion -- we should allow all of `0.12.*`)

pmarks:
Fixes #142

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ default = ["alga"]
 
 [dependencies]
 num-traits = "0.1.32"
-ndarray = "0.11.2"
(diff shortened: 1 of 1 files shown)


vbarrielle:
Looks good, thanks!

PR #147, opened 2019-01-10
Make mul_acc_mat_vec_cs{r|c} more generic

tomtung:


src/sparse/prod.rs
```diff
@@ -54,7 +57,8 @@ pub fn mul_acc_mat_vec_csr<N, I>(
     for (row_ind, vec) in mat.outer_iterator().enumerate() {
         for (col_ind, &value) in vec.iter() {
             // TODO: unsafe access to value? needs bench
-            res_vec[row_ind] = res_vec[row_ind] + in_vec[col_ind] * value;
+            res_vec[row_ind] =
```
tomtung: Wanted to use `dot_dense` here, but since `DenseVector` is passed around by-value, the compiler would (rightly) complain. Maybe we can consider pass-by-ref instead?

vbarrielle: I'm not sure passing by reference is possible here, because it would mean `dot_dense` would not be able to
accept slices...

It's probably ok to either keep the current implementation or take the body from `dot_dense`.

tomtung: Yeah that's right... one workaround could be to change the trait `DenseVector` to something like `DenseVectorRef`, since we don't really need to take ownership of a dense vector anywhere in the code base (at least not yet).

The diff might look like the following:
```diff
diff --git a/src/sparse/prod.rs b/src/sparse/prod.rs
index cac76f8..c5de600 100644
--- a/src/sparse/prod.rs
+++ b/src/sparse/prod.rs
@@ -4,7 +4,7 @@ use num_traits::Num;
 use sparse::compressed::SpMatView;
 ///! Sparse matrix product
 use sparse::prelude::*;
-use sparse::vec::DenseVector;
+use sparse::vec::DenseVectorRef;
 use std::iter::Sum;
 use Ix2;
 
@@ -17,7 +17,7 @@ pub fn mul_acc_mat_vec_csc<N, I, V>(
 ) where
     N: Num + Copy,
     I: SpIndex,
-    V: DenseVector<N>,
+    V: DenseVectorRef<N>,
 {
     let mat = mat.view();
     if mat.cols() != in_vec.dim() || mat.rows() != res_vec.len() {
@@ -45,7 +45,7 @@ pub fn mul_acc_mat_vec_csr<N, I, V>(
 ) where
     N: Num + Copy,
     I: SpIndex,
-    V: DenseVector<N>,
+    V: DenseVectorRef<N>,
 {
     if mat.cols() != in_vec.dim() || mat.rows() != res_vec.len() {
         panic!("Dimension mismatch");
diff --git a/src/sparse/vec.rs b/src/sparse/vec.rs
index 9c23833..be6f1b2 100644
--- a/src/sparse/vec.rs
+++ b/src/sparse/vec.rs
@@ -338,7 +338,7 @@ where
 
 /// A trait for types representing dense vectors, useful for
 /// defining a fast sparse-dense dot product.
-pub trait DenseVector<N> {
+pub trait DenseVectorRef<N> {
     /// The dimension of the vector
     fn dim(&self) -> usize;
 
@@ -350,7 +350,7 @@ pub trait DenseVector<N> {
     fn index(&self, idx: usize) -> &N;
 }
 
-impl<'a, N: 'a> DenseVector<N> for &'a [N] {
+impl<N> DenseVectorRef<N> for &[N] {
     fn dim(&self) -> usize {
         self.len()
     }
@@ -360,7 +360,7 @@ impl<'a, N: 'a> DenseVector<N> for &'a [N] {
     }
 }
 
-impl<N> DenseVector<N> for Vec<N> {
+impl<N> DenseVectorRef<N> for &Vec<N> {
     fn dim(&self) -> usize {
         self.len()
     }
@@ -370,17 +370,7 @@ impl<N> DenseVector<N> for Vec<N> {
     }
 }
 
-impl<'a, N: 'a> DenseVector<N> for &'a Vec<N> {
-    fn dim(&self) -> usize {
-        self.len()
-    }
-
-    fn index(&self, idx: usize) -> &N {
-        &self[idx]
-    }
-}
-
-impl<N, S> DenseVector<N> for ArrayBase<S, Ix1>
+impl<N, S> DenseVectorRef<N> for &ArrayBase<S, Ix1>
 where
     S: ndarray::Data<Elem = N>,
 {
@@ -787,7 +777,7 @@ where
     /// If the dimension of the vectors do not match.
     pub fn dot_dense<T>(&self, rhs: T) -> N
     where
-        T: DenseVector<N>,
+        T: DenseVectorRef<N>,
         N: Num + Copy + Sum,
     {
         assert_eq!(self.dim(), rhs.dim());
@@ -1301,11 +1291,11 @@ mod test {
             assert_eq!(16., vec1.dot_dense(slice));
             assert_eq!(16., vec1.dot_dense(&dense_vec));
         }
-        assert_eq!(16., vec1.dot_dense(dense_vec));
+        assert_eq!(16., vec1.dot_dense(&dense_vec));
 
         let ndarray_vec = Array::linspace(1., 8., 8);
         assert_eq!(16., vec1.dot(&ndarray_vec));
-        assert_eq!(16., vec1.dot_dense(ndarray_vec.view()));
+        assert_eq!(16., vec1.dot_dense(&ndarray_vec.view()));
     }
 
     #[test]
```

vbarrielle: But that would make a breaking change. I think i'd rather explore extending the `DenseVector` trait with an associated type representing a reference type and a method to get it.

tomtung: > extending the `DenseVector` trait with an associated type representing a reference type and a method to get it

Curious how that would work out without changing the API...

vbarrielle:
Rebased and merged manually.

PR #148, opened 2019-01-10
Fix some `cargo clippy` warnings

tomtung:
Fixing the rest would require changing the current public interface (mostly passing refs instead of values when the latter is unnecessary), so leaving them as is for now.

### DIFF
--- a/src/io/mod.rs
+++ b/src/io/mod.rs
@@ -83,7 +83,7 @@ enum DataType {
     Complex,
 }
 
-#[derive(Debug, PartialEq)]
+#[derive(Copy, Clone, Debug, PartialEq)]
 pub enum SymmetryMode {
     General,
     Hermitian,
@@ -151,7 +151,7 @@ where
     loop {
         line.clear();
         let len = reader.read_line(&mut line)?;
-        if len == 0 || line.starts_with("%") {
+        if len == 0 || line.starts_with('%') {
             continue;
         } else {
             break;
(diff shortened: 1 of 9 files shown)


vbarrielle:
That's pretty cool, thanks.

vbarrielle:
About those remaining clippy warnings, such as:

```
error: this argument is passed by value, but not consumed in the function body                                                                                                                                                                                                 
   --> src/sparse/binop.rs:137:10                                                                                                                                                                                                                                              
    |                                                                                                                                                                                                                                                                          
137 |     lhs: CsMatViewI<N, I>,                                                                                                                                                                                                                                               
    |          ^^^^^^^^^^^^^^^^ help: consider taking a reference instead: `&CsMatViewI<N, I>`                                                                                                                                                                                 
    | 
```

I consider this a clippy false positive: passing a view by reference is not ergonomic. This would require a caller to call eg `function(&lhs.view())` instead of `function(lhs.view())`.
Maybe it's possible to silence this warning by implementing `Copy` on `CsMatViewI`?

tomtung:
Yeah I guess we want to add something like
```rust
impl<N, I, IptrStorage, IndStorage, DataStorage> Copy
    for CsMatBase<N, I, IptrStorage, IndStorage, DataStorage>
where
    N: Copy,
    I: SpIndex,
    IptrStorage: Deref<Target = [I]> + Copy,
    IndStorage: Deref<Target = [I]> + Copy,
    DataStorage: Deref<Target = [N]> + Copy,
{
}
```

However, this alone doesn't seem to resolve the warnings. We probably also need to add the `N: Copy`  bound to `struct CsMatBase` itself as well, but that would change the API.

Another alternative (that would also change the API) is to use reference with generics, so that users aren't forced to create view objects when they don't already have one, e.g.,
```diff
-    pub fn convert_mat_storage<N: Clone, I: SpIndex>(
-        mat: CsMatViewI<N, I>,
+    pub fn convert_mat_storage<
+        N: Clone,
+        I: SpIndex,
+        IptrStorage,
+        IndStorage,
+        DataStorage,
+    >(
+        mat: &CsMatBase<N, I, IptrStorage, IndStorage, DataStorage>,
         indptr: &mut [I],
         indices: &mut [I],
         data: &mut [N],
-    ) {
+    ) where
+        IptrStorage: Deref<Target = [I]>,
+        IndStorage: Deref<Target = [I]>,
+        DataStorage: Deref<Target = [N]>,
+    {
         assert_eq!(indptr.len(), mat.inner_dims() + 1);
         assert_eq!(indices.len(), mat.indices().len());
         assert_eq!(data.len(), mat.data().len());
```

Personally I prefer the latter, which is more tedious on the library side, but probably more convenient for users?

PR #150, opened 2019-03-13
Support computing sparse vector dot product using binary search

tomtung:
Using binary search changes the time complexity from O(M+N) to O(MlogN), where M and N are the number of non-zero entries in each vector. This results in a significant speed-up when N>>M.

For more context, I used it here at https://github.com/tomtung/parabel-rs/commit/604970116c69900c12ce12161ea9b3d827961684 where we are doing predictions using linear classifiers trained on sparse vectors with millions of dimensions. The feature vector for a given example is usually sparse with relatively few (e.g. tens of) non-zero entries. The weight vector of a classifier is also quite sparse but still contain many non-zero entries (e.g. 5% of 1 million is still 50,000). As a result, computing the dot product between the two in O(M+N) time is extremely slow.

### DIFF
--- a/src/sparse/prod.rs
+++ b/src/sparse/prod.rs
@@ -7,6 +7,43 @@ use sparse::prelude::*;
 use std::iter::Sum;
 use Ix2;
 
+/// Compute the dot product of two sparse vectors, using binary search to find matching indices.
+///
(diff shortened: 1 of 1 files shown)


vbarrielle:
That's really neat, thanks!

PR #154, opened 2019-03-16
Support computing vector norms

tomtung:
Mostly just porting the logic from [NumPy](https://github.com/numpy/numpy/blob/v1.16.1/numpy/linalg/linalg.py#L2467-L2491) and [Eigen](https://github.com/eigenteam/eigen-git-mirror/blob/3.3.7/Eigen/src/Core/Dot.h#L203-L246).


vbarrielle:
That's perfect, thanks a lot!

Would you be interested in having push access to the repository? It looks like it would be useful to you since sprs is a dependency of one of your projects, and you have contributed excellent work so far.

tomtung:
Thanks, that would be handy :)

PR #157, opened 2019-04-13
implement Dot from linalg

### LINKED ISSUE #156: Implement ndarray::linalg::dot?
I'm trying to write some large-scale SVDs methods that are generic over the type of the input array -- I'd like to to support ndarray::ArrayBase, sprs, and a custom compressed matrix format.  

I noticed that ndarray uses this trait for matrix-matrix or matrix-vector multiply: [ndarray::linalg::Dot](https://docs.rs/ndarray/0.12.1/ndarray/linalg/trait.Dot.html),  but sprs use std::ops::Mul.  (ndarray uses Mul for point-wise multiplication).

1. Would you be open to adding implementations of Dot to mirror the implementations of Mul?
2. Should sprs consider switching to the case convention as ndarray?

pmarks:
fixes #156

### DIFF
--- a/src/sparse/csmat.rs
+++ b/src/sparse/csmat.rs
@@ -16,8 +16,8 @@ use std::mem;
 use std::ops::{Add, Deref, DerefMut, Index, IndexMut, Mul, Range, Sub};
 use std::slice::{self, Iter, Windows};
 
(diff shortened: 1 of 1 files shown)


vbarrielle:
Looks good, thanks!

PR #158, opened 2019-04-13
add a method for reading MatrixMarket from an arbitrary io::BufRead

pmarks:
required for reading from compressed files or the network.


vbarrielle:
Thanks!

PR #159, opened 2019-04-13
implement Eq for CsMatBase

pmarks:
I don't see a reason why not?


vbarrielle:
That was just me forgetting to add it. Thanks!

PR #160, opened 2019-04-29
dense.dot(sparse) impl

pmarks:
This will be allowed on stable once rust-lang/rust#55437 is stabilized.   Seem reasonable?


vbarrielle:
That seems mostly reasonable yes, but I'm not sure I like the added genericity on `csr_mulac_dense_rowmaj` and the likes, I fear it may ask too many type annotations. Maybe it could be
better if `Nout` were equal to `N1` by default.

pmarks:
@vbarrielle this now will compile on stable. It's not possible to have default generic types on methods, so I'm not sure how to reduce the genericity of the `csr_mulacc_dense_rowmaj` methods.

Any thoughts on what we could do there?

vbarrielle:
Hello, sorry for the delayed answer.

I think this added genericity is not too much of a problem, I however need to write a few examples to check if it's actually a breaking change for "typical" calls to `csr_mulacc_dense_rowmaj` (though I dont think it's called that much alone in the wild).

I hope I'll find the time to do it in the coming week.

vbarrielle:
Hello @pmarks, I've rebased and updated your branch on the latest master, and merged it manually, I'm no longer worried about a possible breaking change as I'm going to bump to 0.8 for other changes as well.

I think I got the gist of your PR right, and tests look ok, but please tell me if I forgot something. After rebasing, I obtained the commits https://github.com/vbarrielle/sprs/commit/896c7f2573df200e377a886fba7c5f664db2cbe7 and https://github.com/vbarrielle/sprs/commit/10f42f0103d8ee938d1bcde1a22b986bcc3a3e27

pmarks:
@vbarrielle very cool, thanks!!

PR #161, opened 2019-05-31
allow CsMat::map to return a new type

pmarks:
allow CsMat::map to return a new type


vbarrielle:
Yes it's quite nice to have that, thanks for this fix.

PR #162, opened 2019-05-31
fix borrow warning on nightly

pmarks:
this line is giving the following warning on the nightly compiler. This patch fixes the warning:
```
warning[E0502]: cannot borrow `indptr` as immutable because it is also borrowed as mutable
   --> src/sparse/triplet_iter.rs:122:26
    |
122 |             indptr[i] += indptr[i - 1];
    |             -------------^^^^^^-------
    |             |            |
    |             |            immutable borrow occurs here
    |             mutable borrow occurs here
    |             mutable borrow later used here
    |
    = warning: this error has been downgraded to a warning for backwards compatibility with previous releases
    = warning: this represents potential undefined behavior in your code and this warning will become a hard error in the future
```

### DIFF
--- a/src/sparse/triplet_iter.rs
+++ b/src/sparse/triplet_iter.rs
@@ -119,7 +119,7 @@ where
         let mut indptr = row_counts.clone();
         // cum sum
         for i in 1..=self.rows() {
-            indptr[i] += indptr[i - 1];
+            indptr[i] = indptr[i] + indptr[i - 1];
         }
         let nnz_max = indptr[self.rows()].index();
         let mut indices = vec![I::zero(); nnz_max];


vbarrielle:
Yes I've also had this issue in my branch porting to the 2018 edition. This seems related to order of evaluation that's not the same for a generic index type and for a primitive type https://internals.rust-lang.org/t/rust-expression-order-of-evaluation/2605/30

This fix is fine. Just wondering why the nightly CI doesn't pass, I thought the `dyn` keyword wasn't there until the crate was on the 2018 edition.

vbarrielle:
So it turns out the `dyn` keyword was also stabilized in the 2015 edition, so I've pushed a fixed and I'm rebasing and merging your changes. Thanks!

PR #163, opened 2019-06-15
faster trimat to csmat conversion

pmarks:
Faster method for converting TriMat --> CsMat. Avoids the scanning over the whole row on each element to find duplicates which results in n^2-ish behavior.  Requires slightly more memory than the previous impl (an extra vector of I values).

On a test matrix of size 112137 x 11727 (nnz = 62147198), this change reduces the conversion time from 134s to 8s.

Sample matrix available here:
http://cf.10xgenomics.com/samples/cell-exp/3.0.2/10k_hgmm_v3_nextgem/10k_hgmm_v3_nextgem_filtered_feature_bc_matrix.tar.gz


pmarks:
@vbarrielle any thoughts on this?  Looks like the travis failure is spurious now that I did `cargo fmt`

vbarrielle:
Sorry for answering so late, I've been away a little while. I've relaunched the travis build and it passes now, so it was spurious indeed.

That looks very nice, thank you.

PR #164, opened 2019-07-03
update some old deps

pmarks:


rth:
Thanks @pmarks, LGTM, except for a linting issue. Could you please run `cargo fmt`?

This update is also necessary to get rid of the deprecated `rustc_serialize` dependency (which is problematic for the wasm target https://github.com/rustwasm/team/issues/96)  pulled in older versions of `alga` and `num-complex` crates.

pmarks:
@rth ok fmt'ed now

vbarrielle:
Rebased and merged manually.

PR #166, opened 2019-08-08
Support having different types for indptr & indices in CsMatBase

tomtung:
(Not pushing directly to main branch given it's a rather large diff.)

This diff adds support to set different types for `indptr` & `indices` in `CsMatBase`, which are currently assumed to have the same type. This can be overly restrictive in some cases: even if the indices are small enough to fit into small integer type, the total number of non-zero entries can still cause overflow.

For example, one might want to use `u32` instead of `usize` for indexing to save space. Say each row has 1000 non-zero elements on average, the total number of elements would overflow `u32` with less than 5MM rows.

Since this diff sets the type for `indptr` to be the same as `indices` by default, this shouldn't break most 3rd party dependent code.

### DIFF
--- a/src/sparse/binop.rs
+++ b/src/sparse/binop.rs
@@ -14,53 +14,57 @@ use Ix2;
 use SpRes;
 
 /// Sparse matrix addition, with matrices sharing the same storage type
-pub fn add_mat_same_storage<N, I, Mat1, Mat2>(
+pub fn add_mat_same_storage<N, I, Iptr, Mat1, Mat2>(
     lhs: &Mat1,
     rhs: &Mat2,
-) -> CsMatI<N, I>
+) -> CsMatI<N, I, Iptr>
 where
     N: Num + Copy,
     I: SpIndex,
-    Mat1: SpMatView<N, I>,
-    Mat2: SpMatView<N, I>,
+    Iptr: SpIndex,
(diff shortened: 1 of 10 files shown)


tomtung:
^ build errors seems to be spurious: conneting Github seems to be timing out

rth:
> (Not pushing directly to main branch given it's a rather large diff.)

For people following this repo it would be nice if this was the case also for smaller diffs (excluding trivial fixes) :)

For the record, in `scipy.sparse` in Python both `indptr` and `indices` are also of the same dtype. It is i32 by default, but is upcast to i64 when there are more than elements than `i32::MAX`. In my experience overflow in indptr is most likely. I'm not sure if supporting say u16  and smaller makes sense for indices as that would correspond to less than 33k non zero elements.

Overall this would be quite useful.

tomtung:
> For people following this repo it would be nice if this was the case also for smaller diffs (excluding trivial fixes) :)

Agree, although in this case many parts of the diff are interdependent, making it a little awkward to split up the diff into meaningful sub-parts that each still compiles :( Fortunely the changes in the diff are fairly mechanical and mostly identical to each other, so I hope it's not too hard to read.

tomtung:
Thanks! Looking forward to the next release :)

PR #167, opened 2019-08-08
Make casting from & to SpIndex panic if it overflows

tomtung:
(Not pushing directly to master branch since this diff has performance implications.)

Got bitten by this issue recently (see https://github.com/tomtung/parabel-rs/commit/597a9f2c87a9ac10c33ca31689d60e26d11d6d15), so I think maybe it better to panic even in release mode if arithmatic overflow happens, while providing the option of unchecked casting explicitly.

I think we can look at existing usage in the code base, and figure out where it's safe to do unchecked casts for better performance?

### DIFF
--- a/src/indexing.rs
+++ b/src/indexing.rs
@@ -16,21 +16,25 @@ use num_traits::int::PrimInt;
 /// This is a convenience trait to enable using various integer sizes for sparse
 /// matrix indices.
 pub trait SpIndex: Debug + PrimInt + AddAssign<Self> + Default {
-    /// Convert to usize
+    /// Convert to usize.
     ///
     /// # Panics
     ///
     /// If the integer cannot be represented as an `usize`, eg negative numbers.
-    /// The panic happens in debug builds only.
     fn index(self) -> usize;
 
-    /// Convert from usize
+    /// Convert to usize without checking for overflows.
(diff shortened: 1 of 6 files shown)


src/indexing.rs
```diff
@@ -39,57 +43,59 @@ impl SpIndex for usize {
...
-sp_index_signed_impl!(i64);
-sp_index_signed_impl!(i32);
-sp_index_signed_impl!(i16);
 
-macro_rules! sp_index_unsigned_impl {
-    ($int: ident) => {
-        impl SpIndex for $int {
             #[inline(always)]
-            fn index(self) -> usize {
-                self as usize
```
tomtung: btw since this macro is called on `u64`, it seems to assume `usize` to be at least 64 bits, which is not true for 32 bit builds?

tomtung:
Took a closer look; given the contracts the structs' fields are expected to satisfy, seems like we can do unchecked casts safely in many cases. This can help minimize the overhead.

src/indexing.rs
vbarrielle: Yes I've overlooked this issue when implementing the `SpIndex` trait it seems. Looks like with your new overflow check this issue should be gone.

PR #171, opened 2019-09-16
Fix triplet to CSR conversion with empty input.

### LINKED ISSUE #170: to_csr() can create matrices with invalid indptr
It seems that the conversion function from TrimatI to CsMat does not work properly when the matrix has empty lines. The following code

```rust
let tri_mat = sprs::TriMatI::new((2,4));
let m: sprs::CsMat<u64> = tri_mat.to_csr();
println!("{:?}", m);
let _ = m.get(1,1); 
```

outputs 

```CsMatBase { storage: CSR, nrows: 2, ncols: 4, indptr: [0, 1], indices: [0], data: [0] }```

and then logically panics with `index out of bounds: the len is 2 but the index is 2` in ...`/sprs-0.6.5/src/sparse/csmat.rs:1056:20`.

### PR DESCRIPTION
**Pull request by vbarrielle**
The invariant that, after suppressing duplicates, `slot` should be equal to the number of nonzeros
was not respected with an empty input.

Fixes #170

### DIFF
--- a/src/sparse/triplet.rs
+++ b/src/sparse/triplet.rs
@@ -562,4 +562,14 @@ mod test {
         let csr = triplet_mat.to_csr();
         assert_eq!(csr, expected.to_csr());
     }
+
+    #[test]
+    fn triplet_empy_lines() {
+        // regression test for https://github.com/vbarrielle/sprs/issues/170
+        let tri_mat = TriMatI::new((2, 4));
+        let m: CsMat<u64> = tri_mat.to_csr();
+        assert_eq!(m.indptr(), &[0, 0]);
+        assert_eq!(m.indices(), &[]);
+        assert_eq!(m.data(), &[]);
+    }
 }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/sparse/triplet.rs [hunk 1]
```diff
@@ -562,4 +562,14 @@ mod test {
         let csr = triplet_mat.to_csr();
         assert_eq!(csr, expected.to_csr());
     }
+
+    #[test]
+    fn triplet_empy_lines() {
+        // regression test for https://github.com/vbarrielle/sprs/issues/170
+        let tri_mat = TriMatI::new((2, 4));
+        let m: CsMat<u64> = tri_mat.to_csr();
+        assert_eq!(m.indptr(), &[0, 0]);
```
avangogo: I think m.indptr() must have size outer_dimension+1, so here 3.

vbarrielle: You're completely right, I messed up my invariants.

vbarrielle: For your information I've issued a fix in #172.

PR #172, opened 2019-09-17
Fix handling of empty lines in triplet to CSR/CSC

### LINKED ISSUE #170: to_csr() can create matrices with invalid indptr

vbarrielle:
The previous fix was incomplete due to a bogus test. This commit adds a
more comprehensive test that should cover all edge cases.

Fixes #170 (again...).

### DIFF
--- a/src/sparse/triplet.rs
+++ b/src/sparse/triplet.rs
@@ -564,7 +564,7 @@ mod test {
     }
 
     #[test]
-    fn triplet_empy_lines() {
+    fn triplet_empty_lines() {
         // regression test for https://github.com/vbarrielle/sprs/issues/170
         let tri_mat = TriMatI::new((2, 4));
         let m: CsMat<u64> = tri_mat.to_csr();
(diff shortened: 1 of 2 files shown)


vbarrielle:
Since I've messed this fix the first time, I'll let anyone interested have a quick review in the next few days.

avangogo:
It still doesn't handle correctly the case where the last column/line is empty.
I suggested a fix to the fix (of the fix).

PR #173, opened 2019-09-17
Fix triplet to csr conversion for input ending with empty lines

avangogo:
The previous fix was not working for matrice ending with empty lines/columns.
This commit fixes this issue and adds a related test.


vbarrielle:
Thanks a lot for finding this other edge case and fixing it. This seems much more robust now.

PR #175, opened 2019-10-11
Extend supported ndarray version to 0.13

rth:
As far as I can tell, sprs builds fine and tests pass with ndarray 0.13. 

So I imagine the supported range of versions could be extended (otherwise I can't use it alongside ndarray 0.13).

Not sure if it's necessary to test it in CI (or how to do that).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ default = ["alga"]
 
 [dependencies]
 num-traits = "0.1.32"
-ndarray = ">=0.11, <0.13"
+ndarray = ">=0.11, <0.14"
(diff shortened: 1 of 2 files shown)


rth:
In particular, without support of ndarray 0.13 it will not be possible to build sprs for the WebAssembly target, as previous ndarray versions use `rustc-serialize` which is not supported https://github.com/rustwasm/team/issues/96 (not saying it's the only blocker -- investigating that now).

vbarrielle:
The CI did test with ndarray version 0.13. Since sprs is only using a very tiny portion of the ndarray API, which is exercised in the tests, I think everything is good.

Thanks!

PR #177, opened 2019-10-13
Make serde and serde_derive dependencies optional 

rth:
Currently serde (and serde_derive) are mandatory dependencies, while some users may not need serialization capability. They pull in additional 11 dependencies (e.g. while building sprs), which is non negligible particularly for downstream projects.

I think it would be better to make them optional, as for instance ndarray is doing.

Currently enabling serialization for sprs could be done with,
```
cargo build --features=serde,serde_derive
```

Alternatively those could be put behind a single feature flag (e.g. `serialization` or `serde-1`). 

It is also possible to add them to the list of default features, though I think consistency with `ndarray` on this would be preferable (i.e. not having them enabled by default).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ script:
     - rustc --version
     - cargo --version
     - cargo test --all --verbose
+    - cargo test --features=serde,serde_derive --all --verbose
     - cargo test --release --all --verbose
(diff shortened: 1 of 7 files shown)


vbarrielle:
That's a good idea yes, thanks!

PR #183, opened 2020-02-19
add missing `I` param to the Sub impl for CsVecBase

ztlpn:
Hey, I've noticed that I can't subtract CsVecs with custom index types. This is probably just an omission.

### DIFF
--- a/src/sparse/vec.rs
+++ b/src/sparse/vec.rs
@@ -1185,18 +1185,19 @@ where
     }
 }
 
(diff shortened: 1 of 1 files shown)


tomtung:
Good catch! I must've missed it when I introduced the type parameter.

@vbarrielle I think this shouldn't break any existing code, and doesn't count as a breaking change?

vbarrielle:
Good catch indeed. @tomtung I think you're right, this should not count as a breaking change.

Thanks!

Issue #185, opened 2020-03-26
Profile reduction for cholesky factorizations


LukiRe:
Hey again @vbarrielle, I thought we might continue this discussion in a new issue. 
I am a Student, I research profile reduction methods for a minor assignment, though mostly for my own personal joy. So I have to disappoint you if you believed me to be an expert on the topic.

I will fork the project to implement my version of the RCM algorithm, using the George-Liu pseudoperipheral vertex finder. I will ask any questions that may arise in this issue.

Here is the list of reading material you asked for. I have not read all of these entirely, for that matter, I don't think many humans have ever done that.

- [This](http://heath.cs.illinois.edu/courses/cs598mh/george_liu.pdf) is the essential book describing classsical methods for profile and bandwidth reduction. It describes methods for finding pseudoperipheral verices from page 70 onwards. It also describes the minimum degree reordering, which is generally superior for the fill-in reduction of Cholesky factorizations, from page 133 onward. I might also get around to implementing this ordering.

- [This](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.470.1054&rep=rep1&type=pdf) is the technical report for the implementation of sparse matrices in matlab. It esssentially claims: "Our implementation of symrcm is very similar to the one implemented in SPARSPAK", along with some more general, very interesting remarks. SPARSPAK is the library described in the above book.

- Though only tangential to this conversation, [this](https://link.springer.com/article/10.1007/s40314-016-0394-9?shared-article-renderer) is a recent paper reviewing a variety of different more modern methods, including FNCHC , GPS and NSloan.

One remark about starting vertices: While they (of course) barely matter for random matrices, their choice becomes increasingly important for problems with inherent structure. Pseudoperipheral verices have proven to be superior to simply minimum degree vertices in those cases, as described in the first linked book.

lksriemer:
By the way, just to show a life sign after his little while: I have not forgotten about this project, and intend to implement AMD (or at least support you at implementing it) some time in the future. These last few months, and the upcoming weeks, I have been and will be busy with other matters though.

vbarrielle:
Glad to hear it! Meanwhile, I've started [binding the CAMD part of suitesparse](https://github.com/vbarrielle/sprs/tree/bind_suitesparse_camd), and I'll probably do AMD next, so this should be nice to compare!

PR #186, opened 2020-03-28
Implementation reverse Cuthill-McKee

LukiRe:
See also #185 .
This changes the Cuthill-McKee algorithm implemented here to the reverse Cuthill-McKee algorithm. Starting vertices are choosen by a George-Liu pseudoperipheral vertex finder.
I have added two more test cases and verified the correctness of the exisiting one w.r.t. the new algorithm.

I did try to write in the style of the library, however I might not always have got it right when to use `.index()`, when to call `.unwrap()` vs `.expect()`, and so on. Feel free to change these details, as well as remove unneccesary comments or line breaks. I just can't get enough of those.

As benchmark possibilities are scarce, I did not optimize anything to aggressively. I really believe this crate should come with some sparse matrix collection as well as random matrices to benchmark against, automatically. That would make tweaking the details worthwile. That's an issue for another time though.

Concerning the minimum degree ordering: As it is significantly more complex, It will take some more time to implement it, as in a few days to weeks. I will do it eventually, but I don't want to block somebody else on it, if there is interest.

### DIFF
--- a/src/sparse/linalg/mod.rs
+++ b/src/sparse/linalg/mod.rs
@@ -6,11 +6,10 @@ use num_traits::Num;
 use std::iter::IntoIterator;
 
 pub mod etree;
-mod ordering;
+pub mod ordering;
 pub mod trisolve;
 
 pub use self::ordering::reverse_cuthill_mckee;
-pub use self::ordering::cuthill_mckee_custom;
 
 /// Diagonal solve
 pub fn diag_solve<'a, N, I1, I2>(diag: I1, x: I2)
--- a/src/sparse/linalg/mod.rs
+++ b/src/sparse/linalg/mod.rs
@@ -10,6 +10,7 @@ mod ordering;
 pub mod trisolve;
 
 pub use self::ordering::reverse_cuthill_mckee;
+pub use self::ordering::cuthill_mckee_custom;
 
 /// Diagonal solve
 pub fn diag_solve<'a, N, I1, I2>(diag: I1, x: I2)
--- a/src/sparse/linalg/mod.rs
+++ b/src/sparse/linalg/mod.rs
@@ -9,7 +9,7 @@ pub mod etree;
 mod ordering;
 pub mod trisolve;
 
-pub use self::ordering::cuthill_mckee;
+pub use self::ordering::reverse_cuthill_mckee;
 
 /// Diagonal solve
 pub fn diag_solve<'a, N, I1, I2>(diag: I1, x: I2)
--- a/src/sparse/linalg/ordering.rs
+++ b/src/sparse/linalg/ordering.rs
@@ -13,7 +13,9 @@ pub struct Ordering<I> {
     pub connected_parts: Vec<usize>,
 }
 
-pub fn cuthill_mckee<N, I, Iptr>(mat: CsMatViewI<N, I, Iptr>) -> Ordering<I>
+pub fn reverse_cuthill_mckee<N, I, Iptr>(
+    mat: CsMatViewI<N, I, Iptr>,
+) -> Ordering<I>
 where
     N: PartialEq,
     I: SpIndex,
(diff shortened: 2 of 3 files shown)


vbarrielle:
That looks nice, thanks.

First remarks:
- I think it would be better to expose both `cuthil_mckee` and `reverse_cuthill_mckee`
- could you make sure the comments are wrapped at the 80 characters limit? Running `cargo fmt` should do it automatically. (I'm wondering why no CI launched on your PR?)

I'll take more time to review the rest a bit later.

vbarrielle:
My bad, looks like CI did pass https://travis-ci.org/github/vbarrielle/sprs/builds/668095347, I don't know why github did not display it...

LukiRe:
I am wondering why CI was not displayed as well, but I can't help you there, I am not that good with Github.

- Exposing `cuthill_mckee` too ... I don't know.  It is easily simulated by just reversing the reverse order. Sure, that is definitely not optimal from a performance standpoint, but it shouldn't matter too much, as, firstly, the normal ordering should only be used rarely, and secondly, those O(n) algorithms probably aren't bottlenecks anyway. Still, if you really want it too, that's totally fine by me.

- I contemplated if maybe the algorithm should be generic over a `StartingStrategy`. Then the consumer of the algorithm could decide on how to pick a starting node. Some options that immediately come to mind: `PseudoPeripher, MinimumDegree, Next, Constant`. Being able to optimally choose depending on the specific situation seems like a strong point to me. What do you think?

- I did run `cargo fmt`, no idea why these style guidelines were not enforced. Seems like a problem with the project configuration. I am using VS Code on Windows, even manually running rustfmt with passed `--config max_width=80` option didn't format the comments correctly. Any idea what to do?

Take your time to review, though the tests seem to pass I am not sure if I got all the library-specifics absolutely correct.

vbarrielle:
Don't worry about the CI, maybe it's simply an issue with github.

The main reason why I wanted to expose `cuthill_mckee` as well was to have an API exposing all the choices. It's quite a simple operation indeed, but i feel it's nice to present the user with all the possibilities.

In that spirit, I think it's a idea to expose the `StartingStrategy`, especially if the `StartingStrategy` is made to implement `Default`.

Don't worry about the comments line width, I'll fix them myself later on, and I'll try and find the proper rustmf configuration.

src/sparse/linalg/ordering.rs
```diff
@@ -13,74 +13,223 @@ pub struct Ordering<I> {
...
+    connected_parts.push(nb_vertices);
+
     Ordering {
         perm: PermOwnedI::new(perm),
-        connected_parts,
+        connected_parts: connected_parts
+            .into_iter()
+            .map(|i| nb_vertices - i)
+            .rev()
+            .collect(),
```
vbarrielle: I'm wondering if this could not be done in place here, something like:

```rust
conected_parts.iter_mut().map(|i| *i = nb_vertices - *i);
coneccted_parts.reverse();
```

src/sparse/linalg/ordering.rs
```diff
@@ -13,74 +13,223 @@ pub struct Ordering<I> {
...
+    // Missed optimization: Work with MaybeUninit here.
+    let mut perm = vec![I::default(); nb_vertices];
+
+    for perm_index in 0..nb_vertices {
+        // Find the next index to process, choosing a new starting vertex if necessary.
+        let current_vertex = deque.pop_front().unwrap_or_else(|| {
+            // We found a new connected component. This number will be reverse-inverted later.
+            connected_parts.push(perm_index);
+            // Employ the George-Liu pseudoperipheral vertex finder to find a new starting vertex.
+            find_pseudoperipheral_vertex(&visited, &degrees, &mat).unwrap()
```
vbarrielle: I'd like to be sure this unwrap cannot fail. It looks like `find_pseudoperipheral_vertex` can only fail if all vertices are visited, so it looks like this is panic free. If that is indeed the case, I'd like to see a comment stating it.

src/sparse/linalg/ordering.rs
```diff
@@ -13,74 +13,223 @@ pub struct Ordering<I> {
...
+
+    // Isolated vertices are by definition pseudoperipheral.
+    if degrees[current] == 0 {
+        return Some(current);
+    }
+
+    let (mut contender, mut current_height) =
+        rls_contender_and_height(current, degrees, mat);
+
+    loop {
```
vbarrielle: Could you add comments explaining why this loop should terminate?

src/sparse/linalg/ordering.rs
```diff
@@ -13,74 +13,223 @@ pub struct Ordering<I> {
...
+
+    // For calculating the height.
+    let mut height = 0;
+    let mut current_level_countdown = 1;
+    let mut next_level_countup = 0;
+
+    // The last levels len is used to compute the contender in the end.
+    let mut last_level_len = 1;
+
+    while rls_index < rls.len() {
```
vbarrielle: Isn't this actually a for loop `for rls_index in 0..rls.len()` ? Or maybe I'm missing something.

vbarrielle: My bad, I missed that `rls` was pushed to during the loop.

src/sparse/linalg/ordering.rs
LukiRe: Yes, of course. I made this change, replacing the `map` with a `for_each` though ;-)

src/sparse/linalg/ordering.rs
LukiRe: That is correct. The `unwrap_or_else` can only be triggered if there is still an undiscovered component. This means `find_pseudoperipheral_vertex` can not fail, and this `unwrap` not be triggered.

src/sparse/linalg/ordering.rs
LukiRe: A convincing, more or less formal argument is that, for the loop to continue spinning, the height of the rls of the currently selected vertex must always strictly increase. However, in any connected commponent there is always a set of vertices whose rls's height is maximal. One of those vertices must be reached after a finite amount of iterations, resulting in the loops termination.
I guess I will just comment "no same vertex can be selected as `current` twice", as that's the essential argument.

vbarrielle: Yes I'm ok with this.

src/sparse/linalg/ordering.rs
LukiRe: I will have to work on the error handling here anyway, custom strategies probably can't be trusted like this.

LukiRe:
I will push the requested changes now.

After giving it some more thought, I believe the best compromise between ease of use and customizability is: exposing one function generic over the orderings direction and the starting strategy, `cuthill_mckee_custom(mat, strategy, direction)`, along with one function `reverse_cuthill_mckee(mat)`, which essentialy calls the custom function with the defaults as specified in this PR. I will take a day or so to push this customizable algorithm. Don't merge yet ;-)

vbarrielle:
I agree exposing two functions is the best way to enable choosing the init strategy while encouraging pseudoperipheral as a default.

LukiRe:
Here is the refactored, extended version.
I decided to favour traits for the options. For the strategies that is the obvious choice, for the orderings direction not so much. However I do not see an other way to create this abstraction without considerable runtime costs.

Concerning the namespacing, that definitely needs to be overhauled. I introduced the modules `start` and `order`, but they need to be publicly re-exported in some sensible way or something. It's up to you to decide if and how to do that. Also, maybe a different file structure is needed.

Please comment what you would like to be clarfied etc., and if you believe there need to be more test cases covering all the different options, although most are quite trivial.

src/sparse/linalg/ordering.rs
```diff
@@ -13,74 +11,512 @@ pub struct Ordering<I> {
...
+        /// Contract: This function must always be called with at least one unvisited vertex left.
+        fn find_start_vertex(
+            &mut self,
+            visited: &[bool],
+            degrees: &[usize],
+            mat: &CsMatViewI<N, I, Iptr>,
+        ) -> usize;
+    }
+
+    /// This strategy chooses predetermined starting vertices.
```
vbarrielle: Is this predetermined strategy something common? As a first thought I find it hard to find a sequence of vertices that should work. Or maybe that's a good strategy when one knows very well the sparse structure? In that case, maybe a test case showing a relevant example would be good. Not a priority if that's a lot of work though.

src/sparse/linalg/ordering.rs
```diff
@@ -13,74 +11,512 @@ pub struct Ordering<I> {
...
-                min_deg = vert_deg;
-                min_deg_vert = vert;
+            // Find a new starting vertex, using the given strategy.
+            let new_start_vertex = starting_strategy.find_start_vertex(&visited, &degrees, &mat);
+            assert!(!visited[new_start_vertex], "Vertex returned by starting strategy should always be unvisited");
+
+            new_start_vertex
+        });
+
+        // Write the next permutation in reverse order.
```
vbarrielle: This comment needs to be updated I guess now that the order is custom.

src/sparse/linalg/mod.rs
```diff
@@ -9,7 +9,8 @@ pub mod etree;
 mod ordering;
 pub mod trisolve;
 
-pub use self::ordering::cuthill_mckee;
+pub use self::ordering::reverse_cuthill_mckee;
+pub use self::ordering::cuthill_mckee_custom;
```
vbarrielle: As you mentionned:

> Concerning the namespacing, that definitely needs to be overhauled. I introduced the modules start and order, but they need to be publicly re-exported in some sensible way or something. It's up to you to decide if and how to do that. Also, maybe a different file structure is needed.

without re-exporting the traits and their implementors `cuthil_mckee_custom` is not callable. I'm starting to think the best solution could be to make the `ordering` module public, keeping only the `pub use self::ordering::reverse_cuthill_mckee;` re-export to guide towards the default algorithm.

src/sparse/linalg/ordering.rs
LukiRe: Acctually, it probably is not. It is definitely something which *could* be of interest, only to a small subset of problems though, when somebody knows the structure of the matrix very well. For that case we would then, however, have to enable choosing multiple starting vertices as well.
That does seem like to much. I will remove this strategy, pointing out in the trait documentation custom strategies can be implemented by the user, with a note that we would love to be informed about what custom strategies users implement.

src/sparse/linalg/ordering.rs
LukiRe: Yes.

src/sparse/linalg/mod.rs
LukiRe: Will do that as described.

LukiRe:
I have implemented all changes as requested. If there is nothing left to dicuss, this can now be merged.

vbarrielle:
That's perfect, I'm merging.

vbarrielle:
ANd thanks a lot!

LukiRe:
> 
> 
> And thanks a lot!

It's the magic of open source!

Issue #184, opened 2020-03-04
Quadratic complexity of sparse matrices multiplication


espdev:
Hello!

I try to use sprs package in my experimental code. I have encountered the issue of very slow multiplication of large sparse matrices.

My code uses large diagonal matrices in CSR format. The matrices might contain from hundreds/thousands to millions `NNZ` elements. Multiplication of such matrices in sprs has almost quadratic complexity `O(nnz^2)`. So we cannot use sprs in production code currently.

I wrote some tests in Rust+sprs and Python+scipy. SciPy [uses](https://github.com/scipy/scipy/blob/master/scipy/sparse/sparsetools/csr.h#L514) SMMP algorithm ([Sparse Matrix Multiplication Package](https://www.i2m.univ-amu.fr/perso/abdallah.bradji/multp_sparse.pdf)) and perform multiplication such matrices in tens/hundreds of milliseconds.

The code of my benchmarks can be found on gist: https://gist.github.com/espdev/d278962828ce7360d653b8679d7bb511

The plot for `mat * mat.T`:
![sprs_scipy_matmul](https://user-images.githubusercontent.com/1299189/75873314-c2e56780-5e20-11ea-929b-abce8e1519b6.png)

Could you consider using SMMP algorithm in sprs? This algorithm has complexity `O(n_row*K^2 + max(n_row,n_col))` where K is the maximum nnz in a row of A and column of B.

rth:
Thanks for investigating this! I have always wondered how sprs would compare to scipy.sparse performance wise https://github.com/scipy/scipy/issues/10201#issuecomment-582426961 It would be interesting to also have this comparison random sparse arrays to evaluate the impact of shape and density.

vbarrielle:
That's a problem indeed, thanks for reporting it. I'll try and find some time to investigate.

espdev:
I have compared sprs and scipy on random sparse matrices with some shapes and density. It is interesting. It seems I incorrectly defined the dependency sprs complexity from nnz. sprs mat*mat complexity directly depends on matrices sizes also. The performance is dramatically reduced while increasing the sizes of the matrices independently from nnz. At the same time in scipy the dependency is quite different. 

Here are some plots from my benchmarks.

**scipy**:
![scipy_sparse_matmul](https://user-images.githubusercontent.com/1299189/75931902-328e3d80-5e87-11ea-94b4-032d8fa619e9.png)

**sprs**
[...]

PR #187, opened 2020-04-18
Better performance for sparse matrix product

vbarrielle:
As shown in #184, the current implementation of sparse matrix product had quadratic behavior. As suggested, this branch implements the paper [Sparse Matrix Multiplication Package](https://www.i2m.univ-amu.fr/perso/abdallah.bradji/multp_sparse.pdf) to get better performance.

Benchmarks are given in #184.


espdev:
@vbarrielle 
Great work! Thank you very much!

Issue #190, opened 2020-05-04
One-Hot Matrix constructor


austinjones:
Would you be interested in a one-hot matrix constructor?  I'm working on one based on sprs.

If constructed from a sparse matrix, the constructor would reduce each inner vector to the element with the maximum value.  As a result, along the inner dimension, exactly one element would be non-zero (with value of 1).

If you are interested, let me know and I'll polish it up for a PR.

vbarrielle:
Hello,

That looks nice to have, please go ahead.

Thanks!

PR #191, opened 2020-05-07
Add to_inner_onehot, into_csc, and into_csr

austinjones:
Hey @vbarrielle,
As in #190, here is an implementation of 'to_inner_onehot'.

I thought about writing 'to_row_onehot' and 'to_column_onehot', because users would probably want to onehot along a specific direction, but settled on adding 'into_csc' and 'into_csr', which cheaply return self if the CSC/CSR compression type is correct.  They can only be called on an owned CsMatBase with Vec storage, as the return type between the branches needs to be equal. 
 Users can chain them: `mat.into_csr().to_inner_onehot()`

Also, the where bound could be lowered to `N: Num` if to_inner_onehot panicked on NaNs.  Right now it requires `N: Float` so that NaNs can be filtered out before the max_by.

Let me know if you have any comments.

### DIFF
--- a/src/sparse/csmat.rs
+++ b/src/sparse/csmat.rs
@@ -1,5 +1,5 @@
 use ndarray::ArrayView;
-use num_traits::{Num, Signed, Zero};
+use num_traits::{Float, Num, Signed, Zero};
 use std::cmp;
 ///! A sparse matrix in the Compressed Sparse Row/Column format
 ///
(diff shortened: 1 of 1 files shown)


vbarrielle:
This is neat, thanks! I do agree it's good to chain using `mat.into_csr().to_inner_onehot()`. Restricting to `N: Float` seems reasonable as well, if eg integers are later needed it should be possible to extend the API.

PR #193, opened 2020-06-06
change README to markdown

mulimoen:
I noticed the page on `crates.io` for this crate was not rendering the README properly. It seems `crates.io` unfortunately does not support `.rst`, so I'm submitting this PR to change the markdown to markdown, which will render correcly both on github and on `crates.io`.


vbarrielle:
That's a good change, I originaaly picked rst for the readme but I was regretting this choice, thanks for doing this change.

PR #194, opened 2020-06-08
Documentation and other small updates

mulimoen:
When initially stumbling upon this crate I was a bit confused on how to create a matrix by using indices. I've added an example of how to create a matrix using the `TriMat` structure, which one can later convert to a compressed format.

In addition I updated the dependencies, fixed two doc-links, and some clippy lints.

### DIFF
--- a/examples/fill_in_reduction.rs
+++ b/examples/fill_in_reduction.rs
@@ -5,7 +5,7 @@ extern crate ndarray;
 extern crate sprs;
 extern crate tobj;
 
-use image::{ImageBuffer, ImageError, ImageResult, Luma};
+use image::{ImageBuffer, ImageResult, Luma};
 use ndarray::{arr2, ArrayView2};
 use std::env;
 use std::error::Error;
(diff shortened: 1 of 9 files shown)


examples/fill_in_reduction.rs
```diff
@@ -38,7 +38,7 @@ fn small_lap_mat() -> sprs::CsMat<f64> {
 }
 
 fn lap_mat_from_obj(path: &str) -> Result<sprs::CsMat<f64>, tobj::LoadError> {
-    let (objects, _materials) = tobj::load_obj(&Path::new(path))?;
+    let (objects, _materials) = tobj::load_obj(&Path::new(path), false)?;
```
vbarrielle, line 41: The laplacian matrix is built from triangles, so I think the `triangulate` flag passed to `load_obj` should be set to `true`.

vbarrielle:
This looks nice, I'll have to take some time to think more about this flag in `tobj::load_obj`, and look at the test results.

mulimoen:
The old behaviour seems to default to `true` for this option, https://github.com/Twinklebear/tobj/commit/1ae79e6d1ada57950e4589b32921b28c704e7140

vbarrielle:
Looks good with this change, thanks!

vbarrielle:
This will mean a version bump when I next publish the crate, but I was planning on doing one when I'm done updating the matrix products (see #192 for context)

PR #195, opened 2020-06-09
export iterators

mulimoen:
It seems some structures are not exported in the docs, which makes it hard to understand how they work, and which traits they implement.

It would be nice if such items were exported automatically, which they are if `pub mod sparse` is added to the topleve. But adding `pub mod sparse` will add `Re-exports` to all the type definitions on the toppage, which makes it difficult getting an overview. Is there a better way of handling this?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -106,6 +106,7 @@ pub type Ix2 = ndarray::Ix2;
 pub use indexing::SpIndex;
 
 pub use sparse::{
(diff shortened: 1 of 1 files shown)


vbarrielle:
That looks good, thanks. I don't think `pub mod sparse` would be a good idea, I think this current module hierarchy is an historical choice that's a bit unfortunate, ideally I'd want to have a cleaner module hierarchy. That's probably something that needs to be done in the future, though I'd probably like to convert the crate to the 2018 edition beforehand to benefit from the niceties of the more modern module syntax.

PR #196, opened 2020-06-14
Update to 2018 edition

mulimoen:
Most of these changes are done automatically by `cargo fix --edition`. This is a very mechanistic PR which should not result in any API differences.

There is also a change for serde, where the `serde_derive` crate was previously imported, instead of using the `derive` feature on `serde` directly. This is a breaking change.

In #193 I missed one `rst`, so I've changed this in this PR


vbarrielle:
Nice! However it looks like the CI build is failing https://travis-ci.org/github/vbarrielle/sprs/builds/698178639 (for some reason it's no longer integrated in github, I still haven't figured out why). Looks like you need to run `cargo fmt`.

mulimoen:
I ran `cargo fmt` and fixed a CI failure

Strange with Travis, I'll see if I can cook up a github actions. I've been looking to do so on some of my own repositories.

vbarrielle:
That's perfect, thanks!

PR #198, opened 2020-06-17
Add github actions

mulimoen:
This adds github actions, which are run on every PR and push to master/main, and every month.

The time it takes for CI is reduced significantly, a rustfmt seems to take less than 30 seconds, with the full suite taking about 5 minutes. The crates are now tested on `ubuntu`, `mac`, and `windows`, but the `suitesparse_bindings` are only built on ubuntu.

Some differences between travis and github actions (these can be added if wanted):
- No longer running the heat example
- Tests are not run with release
- Nightly feature is not enabled for `sprs-benches`

If this PR is merged, the actions will automatically be run, without any futher setup.

The workflow is inspired mostly by the implementation in `ripgrep` (license: unlicense)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,107 @@
+name: CI
+
+on:
+  push:
+  pull_request:
+    branches:
+    - master
+  schedule:
+  - cron: '0 0 15 * *'
+jobs:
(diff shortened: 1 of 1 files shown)


vbarrielle:
That's nice, thanks!

Issue #205, opened 2020-08-03
Exposing PrimtiveKind as public ?


k3yavi:
Hi guys,

Thanks a lot for the awesome crate. I am trying to extend the crate for my use case [here](https://github.com/k3yavi/SingleCellExperiment/blob/master/src/mtx.rs#L22).
Since the class I am working on can have generic matrix primary type, I have to extend it using the `sprs::io::write_matrix_market` function . However, the generic matrix to the function expects that it implements the `PrimitiveKind` trait from the `num_kinds` module which is a private module for the library. Is there a work around for this ? Thanks in advance.

the exact error I get is
```
error[E0277]: the trait bound `MatValT: sprs::num_kinds::PrimitiveKind` is not satisfied
   --> src/mtx.rs:22:46
    |
22  |     sprs::io::write_matrix_market(_path_str, _matrix)?;
    |                                              ^^^^^^^ the trait `sprs::num_kinds::PrimitiveKind` is not implemented for `MatValT`
    |
   ::: /Users/k3yavi/.cargo/registry/src/github.com-1ecc6299db9ec823/sprs-0.8.0/src/io.rs:280:13
    |
280 |     N: 'a + PrimitiveKind + Copy + fmt::Display,
    |             ------------- required by this bound in `sprs::io::write_matrix_market`
    |
help: consider further restricting this bound
    |
20  |     MatValT: std::fmt::Display + Copy + sprs::num_kinds::PrimitiveKind,
```

k3yavi:
[this](https://github.com/k3yavi/sprs/commit/517ecc1a1c98e03cc28699cb6cc2ef1fd3d142c4) seems to have fixed it for me, not sure if this is the right thing to do.

vbarrielle:
Hello @k3yavi, it's indeed unfortunate that this trait is private, making it public looks like the right way to fix the issue. If you pull request your fix I'll gladly merge it.

Thanks for the report!

vbarrielle:
@k3yavi just FYI, I've published version 0.8.1 of sprs on crates.io including your changes, so you can use crates.io in your crates if needed.

k3yavi:
Thanks for the heads up @vbarrielle , much appreciated. I'll update the dependency on my end in the next rounds of revision.

PR #206, opened 2020-08-10
Update lib.rs

k3yavi:
A PR to expose the `num_traits` as public, it is also the fix for https://github.com/vbarrielle/sprs/issues/205.


vbarrielle:
Thanks!

PR #210, opened 2020-08-23
impl approx

### LINKED ISSUE #197: Add approx traits
It would be nice if the sparse matrices support the traits from `approx`, which allows checking whether two matrices containing floats are approximately equal. This to aid in testing of code, both internally in this crate and outside.

mulimoen:
This implements `approx` for sparse vectors and matrices. This can be very useful for comparing matrices, as the  `PartialEq` checks for structural equality, and the approx traits "ignore" the storage type.

Fixes #197

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -159,6 +159,15 @@ pub enum FillInReduction {
     ReverseCuthillMcKee,
 }
 
+#[cfg(feature = "approx")]
+/// Traits for comparing vectors and matrices using the approx traits
+///
+/// Comparisons of sparse matrices with different storages might be slow.
+/// It is advised to compare using the same storage order for efficiency
+pub mod approx {
+    pub use approx::{AbsDiffEq, RelativeEq, UlpsEq};
+}
+
 #[cfg(test)]
 mod test_data;
 
--- a/src/sparse/csmat.rs
+++ b/src/sparse/csmat.rs
@@ -3160,10 +3160,10 @@ mod approx_impls {
             m2.add_triplet(1, 1, 8_u8);
             let m2 = m2.to_csr();
 
-            approx::assert_abs_diff_ne!(m1, m2);
-            approx::assert_abs_diff_ne!(m1, m2.to_csc());
-            approx::assert_abs_diff_ne!(m1.to_csc(), m2);
-            approx::assert_abs_diff_ne!(m1.to_csc(), m2.to_csc());
+            ::approx::assert_abs_diff_ne!(m1, m2);
+            ::approx::assert_abs_diff_ne!(m1, m2.to_csc());
+            ::approx::assert_abs_diff_ne!(m1.to_csc(), m2);
+            ::approx::assert_abs_diff_ne!(m1.to_csc(), m2.to_csc());
         }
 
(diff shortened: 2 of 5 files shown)


src/sparse/csmat.rs
```diff
@@ -2953,3 +2953,298 @@ mod test {
...
+            } else {
+                // Checks if all elements in self has a matching element
+                // in other
+                if !self.iter().all(|(n, (i, j))| {
+                    n.abs_diff_eq(
+                        other
+                            .get(i.to_usize().unwrap(), j.to_usize().unwrap())
+                            .unwrap_or(&N::zero()),
+                        epsilon.clone(),
+                    )
```
vbarrielle, line 3005: I'm wondering if all these binary searches are not going to end up quite expensive, maybe in some cases it's best to convert one operand to the other's storage. But I'm not sure, since it may be costly to allocate a new matrix in a comparison.

vbarrielle (approved):
Thanks for this nice implementation.

As I mentionned there may be a performance penalty when the matrices have different storages, but it may be good to keep it that way, the caller can always explicitly convert the input matrix. Maybe the performance characteristics should be documented somewhere, but I'm not sure it's possible to document on a trait implementation.

I'll merge this tomorrow, except if you feel there are some missing bits.

src/sparse/csmat.rs
mulimoen: I tried the cloning approach, but that leads to problems with expected type parameters, seems one can't expect a recursive call to resolve to different parameters

mulimoen: We could add a special case when all the elements in both are coinciding, so we only have to check the matrices one way instead of both

vbarrielle: The more I think about it, the more I think it's not a problem. It's ok for the comparison to be slower when the storages differ, the caller can deal with it by explictly changing storages, we simply need to document it.

mulimoen:
By placing the traits in a module in the toplevel we can squeeze some documentation in. If one clicks the trait implementation `AbsDiffEq` one gets redirected to `sprs::approx::AbsDiffEq`, and `sprs::approx` got the information regarding performance.

Alternatively one can also directly import the traits and add a doc string, but these docs will be above the trait item explanation, which makes it a bit odd...

src/lib.rs
```diff
@@ -159,6 +159,15 @@ pub enum FillInReduction {
     ReverseCuthillMcKee,
 }
 
+#[cfg(feature = "approx")]
+/// Traits for comparing vectors and matrices using the approx traits
+///
+/// Comparisons of sparse matrices with different matrices might be slow.
```
vbarrielle, line 165: > Comparisons of sparse matrices with different matrices might be slow.
I think you meant "with different storages" and not "with different matrices"

mulimoen: Indeed!

mulimoen:
Maybe we should add some [metadata](https://docs.rs/about/metadata) for `docs.rs` to have all features enabled and visible in the docs?

vbarrielle:
Yes that's a good idea, I've filed it in #215.

PR #211, opened 2020-08-24
Add Kronecker product

### LINKED ISSUE #209: Implementing the Kronecker product
I would like to see an implementation of the kronecker product in this library, which can be used for i.e. translating multi-dimensional spaces down to a single dimension.

Currently I use the following:
https://gist.github.com/mulimoen/6321497759ce8ec0a14dcf4bb1bd9706

Let me know if you'd like me to make a pull request!

mulimoen:
This is a slightly modified version of the one in #209, which uses some symmetries for the `CSC^2` -> `CSR^2` case.

The incompatible storage case is now handled by converting the second matrix to the opposite storage. This might not be the most efficient, but the performance should be predictable in this case.

Closes #209

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -93,11 +93,11 @@ pub use crate::indexing::SpIndex;
 
 pub use crate::sparse::{
     csmat::CsIter, csmat::OuterIterator, csmat::OuterIteratorMut,
-    csmat::OuterIteratorPerm, CsMat, CsMatBase, CsMatI, CsMatVecView,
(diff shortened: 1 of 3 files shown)


vbarrielle:
Thanks, that's a very clean implementation.

> This might not be the most efficient, but the performance should be predictable in this case.

Indeed, it might be best to dynamically choose which matrix should be converted, but this would need benchmarks first. I think it's best to first have this implementation, then later we'll see if the performance needs to be improved. And since the performance is predictable, this means the calling code is free to perform its own manipulations to enforce best performance.

PR #212, opened 2020-08-24
Optional rayon

### LINKED ISSUE #208: Make rayon optional
After upgrading to `0.8` the crate can no longer be used on the `wasm32` target, as a threadpool can not be initialized:

> panicked at 'The global thread pool has not been initialized.: ThreadPoolBuildError { kind: IOError(Custom { kind: Other, error: "operation not supported on this platform" }) }'

Could the dependency on rayon be made into a (default) feature instead, to allow usage on this platform?

mulimoen:
As mentioned in #208 rayon will panic when creating a threadpool on targets which do not have threads. This PR puts `rayon` and `num_cpus` behind the feature flag `multi_thread` (suggestions for name?), but otherwise follows exactly the same code path as the `Fixed(1)` case.

Fixes #208

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,3 +109,19 @@ jobs:
     - name: Run benchmarks
       run: |
           cargo bench --workspace
+
+  optional_features:
+    name: Optional features
+    runs-on: ubuntu-18.04
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+    - name: Install rust
+      uses: actions-rs/toolchain@v1
+      with:
+          toolchain: nightly
+          override: true
+          profile: minimal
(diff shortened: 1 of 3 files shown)


vbarrielle:
That's very nice, thanks! The feature name `multi_thread` is good in my opinion.

mulimoen:
This is a breaking change btw, as someone could have `default-features = False` in their `Cargo.toml`

vbarrielle:
I had not realized that. Thanks for pointing it out. Guess I'll try to syncrhonize the version bump with another breaking change I'm thinking of. I should probably maintain a file of planned breaking changes.

vbarrielle:
And eventually find ways to commit to a more stable API of course...

PR #216, opened 2020-09-01
Add #[inline] to indexing

mulimoen:
Adding the `#[inline]` attribute to the indexing operators got a speedup of 15-30% on `mul_acc_mat_vec_csr` for some matrices. The benchmarks in this library does not gain or lose anything from this change.

Further observations: Changing to unsafe access in `mul_acc_mat_vec_csr` does not modify the performance characteristics compared to the safe version.

### DIFF
--- a/src/sparse/vec.rs
+++ b/src/sparse/vec.rs
@@ -361,6 +361,7 @@ impl<'a, N: 'a> DenseVector<N> for &'a [N] {
         self.len()
     }
 
+    #[inline]
     fn index(&self, idx: usize) -> &N {
         &self[idx]
     }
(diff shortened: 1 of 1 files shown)


vbarrielle:
That's a good change, these short functions are good candidates for inlining.

> The benchmarks in this library does not gain or lose anything from this change.

I don't think they test `mul_acc_mat_vec_csr`, what's tested there uses iterators, so this explains why there's no difference.

Thanks for finding this!

mulimoen:
@vbarrielle I would expect all functions using `SparseVector` to benefit from this, including the benched function `dot_dense`. It might be that it is inlined automatically inside the crate, but does not pass across the crate barrier.

vbarrielle:
Yes that's very probable actually.

PR #217, opened 2020-09-05
make some small improvements in the utils:sort_indices_data_slices

cxzheng:
Just improved the utils::sort_indices_data_slices a bit, while I'm using this crate (very nice crate!)
The change includes 1) the removal of the explicit bound check using iterators, and 2) a quick test to verify my changes.

### DIFF
--- a/src/sparse.rs
+++ b/src/sparse.rs
@@ -256,7 +256,6 @@ pub trait SparseMat {
 
 mod utils {
     use crate::indexing::SpIndex;
-    use itertools::izip;
 
     pub fn sort_indices_data_slices<N: Copy, I: SpIndex>(
         indices: &mut [I],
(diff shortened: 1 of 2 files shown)


src/sparse.rs
```diff
@@ -268,15 +269,15 @@ mod utils {
...
+        for (i, v) in indices.iter().zip(data.iter()) {
+            buf.push((*i, *v));
         }
 
         buf.sort_unstable_by_key(|x| x.0);
 
-        for (i, &(ind, x)) in buf.iter().enumerate() {
-            indices[i] = ind;
-            data[i] = x;
+        for (i, v, &mut (ind, x)) in izip!(indices, data, buf) {
```
mulimoen, line 278: We could use
```rust        
for (&(i, x), (ind, v)) in
            buf.iter().zip(indices.iter_mut().zip(data.iter_mut()))
```
to avoid pulling in itertools

cxzheng: I agree it might not worth pulling in itertools just for this small improvement.

vbarrielle:
Thanks for noticing and fixing the issue, @cxzheng ! And thanks @mulimoen for reviewing and pointing out that itertools was not necessary.

PR #218, opened 2020-09-08
fixed a bug in mul_acc_mat_vec_csr and mul_acc_mat_vec_csc

cxzheng:
There is a bug in `mul_acc_mat_vec_csr` and `mul_acc_mat_vec_csc`. Basically, the output `res_vec` needs to be set to zero before accumulating values in the matrix vector multiplication.

I also btw did some small tweaks following the suggestions from `cargo clippy`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -73,6 +73,7 @@ assert_eq!(a, b.to_csc());
 ```
 
 */
+#![deny(warnings)]
 
 pub mod array_backend;
 pub mod errors;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -74,8 +74,6 @@ assert_eq!(a, b.to_csc());
 
 */
 
-#![deny(warnings)]
-
 pub mod array_backend;
 pub mod errors;
 pub mod indexing;
--- a/src/sparse/prod.rs
+++ b/src/sparse/prod.rs
@@ -65,7 +65,6 @@ pub fn mul_acc_mat_vec_csc<N, I, Iptr, V>(
         panic!("Storage mismatch");
     }
 
-    res_vec.iter_mut().for_each(|v| *v = N::zero());
     for (col_ind, vec) in mat.outer_iterator().enumerate() {
         let multiplier = in_vec.index(col_ind);
         for (row_ind, &value) in vec.iter() {
@@ -95,8 +94,10 @@ pub fn mul_acc_mat_vec_csr<N, I, Iptr, V>(
     }
 
     for (row_ind, vec) in mat.outer_iterator().enumerate() {
-        let tv = res_vec.get_mut(row_ind).unwrap(); // this unwrap is ok because we did the check
-        *tv = N::zero();
+        // this unwrap is ok because we did the check before to ensure
+        // mat.row() == res_vec.len() and now the row_ind is within the
+        // range of [0, mat.row). So it should be safe.
+        let tv = res_vec.get_mut(row_ind).unwrap();
         for (col_ind, &value) in vec.iter() {
             // TODO: unsafe access to value? needs bench
             *tv = *tv + *in_vec.index(col_ind) * value;
--- a/src/sparse/prod.rs
+++ b/src/sparse/prod.rs
@@ -65,6 +65,7 @@ pub fn mul_acc_mat_vec_csc<N, I, Iptr, V>(
         panic!("Storage mismatch");
     }
 
+    res_vec.iter_mut().for_each(|v| *v = N::zero());
     for (col_ind, vec) in mat.outer_iterator().enumerate() {
         let multiplier = in_vec.index(col_ind);
         for (row_ind, &value) in vec.iter() {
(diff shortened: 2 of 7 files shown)


vbarrielle:
Hello,

That's not a bug, that's the expected behaviour. The function accumulates in the output vector, so should preserve the contents it has on input. It probably should be stated more explicitly in the docs. Maybe some line like `This function implements the operation y += A * x`.

This accumulation behavior is desirable because it enables adding matrix/vector products with only one storage vector, ie the equivalent of `y = A * x + B * x` is something like

```
let mut y = [0; n];
mul_acc_mat_vec_csr(b.view(), x, &mut y);
mul_acc_mat_vec_csr(a.view(), x, &mut y);
```

One could argue this is not convenient, but this function is quite low-level, the high level interface is the `*`operator. The drawback of this high-level operator is that it allocates, and does not work on slices. Maybe we could add a version of `mul_acc_mat_vec_cs{r,c}` that has this zeroing behavior for that purpose, eg `mul_mat_vec_cs{r,c}`.

The clippy suggested fixes look nice.

src/lib.rs
```diff
@@ -74,8 +74,6 @@ assert_eq!(a, b.to_csc());
 
 */
 
-#![deny(warnings)]
```
vbarrielle, line 77: I've been wondering about removing this for a while, but I'm yet unsure. I've found this switch quite helpful to force me to think about warnings.

I suspect the rationale is that it could break some downstream crates with some compiler changes, which is a very valid concern, I'm wondering if it's possible to have this switch on only for dev builds?

src/sparse/prod.rs
```diff
@@ -94,10 +95,11 @@ pub fn mul_acc_mat_vec_csr<N, I, Iptr, V>(
     }
 
     for (row_ind, vec) in mat.outer_iterator().enumerate() {
+        let tv = res_vec.get_mut(row_ind).unwrap(); // this unwrap is ok because we did the check
```
vbarrielle, line 98: I'm not sure this comment is clear, I think the index is in bounds because `CsMat`guarantees its indices to be in bounds, and because the dimensions have been checked to be in agreement.

It's however a very good thing to move this indexing out of the inner loop, this means the index checking will not be too much of a cost, good catch!

vbarrielle (changes requested):
As I mentionned in my previous message, I think the way forward is to add a non-accumulating version of this API, and to be more precise in the documentation of `mul_acc_mat_vec_cs{r,c}` to state more clearly that accumulation is the desired behavior (and probably point to the new non-accumulating version as an alternative).

src/lib.rs
mulimoen: You can run CI with "RUSTFLAGS=-D warnings" and remove that line. Any PRs with warning should then fail

cxzheng: OK. I put #![deny(warnings)] back, while I personally tend to use `RUSTFLAGS=-D warnings` as @mulimoen suggested. Anyway, it's up to you to decide. :)

cxzheng:
> Hello,
> 
> That's not a bug, that's the expected behaviour. The function accumulates in the output vector, so should preserve the contents it has on input. It probably should be stated more explicitly in the docs. Maybe some line like `This function implements the operation y += A * x`.
> 
> This accumulation behavior is desirable because it enables adding matrix/vector products with only one storage vector, ie the equivalent of `y = A * x + B * x` is something like
> 
> ```
> let mut y = [0; n];
> mul_acc_mat_vec_csr(b.view(), x, &mut y);
> mul_acc_mat_vec_csr(a.view(), x, &mut y);
> ```
> 
> One could argue this is not convenient, but this function is quite low-level, the high level interface is the `*`operator. The drawback of this high-level operator is that it allocates, and does not work on slices. Maybe we could add a version of `mul_acc_mat_vec_cs{r,c}` that has this zeroing behavior for that purpose, eg `mul_mat_vec_cs{r,c}`.
> 
> The clippy suggested fixes look nice.

Sorry for my misunderstanding of the two functions' behaviors. What you explained makes sense. I changed them back.

src/sparse/prod.rs
cxzheng: I added some more comment to explain why I think it's ok to `unwrap` there. My understanding is that the earlier check `mat.rows() != res_vec.len()` in that function ensures the `unwrap` is fine.

src/lib.rs
vbarrielle: Nice to know, thanks @mulimoen

src/sparse/prod.rs
vbarrielle: The new comment is clearer, thanks!

vbarrielle (approved):
Thanks for your quick iteration @cxzheng. Do you want me to add the non-accumulation versions `mul_acc_mat_vec_cs{r,c}`, or are you using the `*` operator?

cxzheng:
> Thanks for your quick iteration @cxzheng. Do you want me to add the non-accumulation versions `mul_acc_mat_vec_cs{r,c}`, or are you using the `*` operator?

It's okay for me. I'm just gonna set the vector to zero before calling `mul_acc_mat_vec_cs{r,c}`. I think it makes sense to stay with the same style like `saxpy` in BLAS in this case.

PR #225, opened 2020-09-18
Minimum rust version

mulimoen:
When I bisected I did not take into account the other crates in the library. I've fixed this on top of your msrv branch.


vbarrielle:
I had not noticed you were working on this at the same time I was on it, sorry!

mulimoen:
You opened your PR five minutes after this (: Only difference is a call to max_value as opposed to INT32_MAX

PR #227, opened 2020-09-19
Implement MulAssign and DivAssign

mulimoen:
This implements `MulAssign` and `DivAssign` for matrices and vectors.

The `MulAssign` could be expressed as `scale`, but it's nice to have multiple options. `DivAssign` might be less efficient than doing inversion of `rhs` then scaling, altough some cases migh benefit from direct division.

The `impl`s are quite broad, but this should not be a problem as sparse inplace vector-vector multiplication is probably not very efficient.

(The first commit removes a deprecated type)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -82,8 +82,6 @@ pub mod num_kinds;
 mod sparse;
 pub mod stack;
 
-/// Deprecated type alias, will be removed on next breaking change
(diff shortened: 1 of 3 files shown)


vbarrielle:
All good, thanks!

Nice catch noticing that deprecated type

Issue #99, opened 2016-10-19
Add CSR/CSC product subroutine


vbarrielle:
Requiring conversion before the product can be really slow. This requires setting up some benchmarks to measure the concrete effect.

maboesanman:
I'm interested in working on this! In particular it seems CSC * CSR is actually broken at the moment, and this seems like a good place to fix that issue, as well as improve performance. Will you accept a PR?

mulimoen:
PRs are accpepted. Would be nice to start with some basic benchmarks to see how we compare against `scipy` and the likes.

vbarrielle:
Beware though, this issue is quite old, and since then matrix products has seen several changes, mostly the SMMP implementation #187 so I think the first thing to do would be to enhance [the existing product benchmarks](https://github.com/vbarrielle/sprs/tree/master/sprs-benches) to see precisely the impact of storage conversion.

You mentionned CSC*CSR being broken, do you have an example?

maboesanman:
I don't have a functional code example handy, but if you take two `CsVec`, call them `a` and `b`, and you do `a.col_view() * b.row_view()` you get a panic, instead of a len(b) x len(a) matrix

PR #229, opened 2020-09-23
Fix (CSC, CSR) multiplication

mulimoen:
Multiplication of `(CSC, CSR)` matrices would lead to a crash for valid matrices. `(CSC, CSR)` should instead copy the format of `(CSC, CSC)` with `rhs` transposed.

Ref #99


vbarrielle:
Looks good, thanks for the quick investigation and fix!

PR #233, opened 2020-09-30
Build suitesparse from source

mulimoen:
This adds feature flags for `suitesparse_ldl_sys` and `suitesparse_camd_sys` to build `SuiteSparse` from source and statically link it. This should be easily extendable to more components later on if more parts are wanted. We might want to add this feature flag to the `sprs_suitesparse` packages.

The build script for `SuiteSparse` reimplements the `Makefile`s, as these were hard to reuse, I've added CI which should ensure we won't break this.

The targeted version of `SuiteSparse` is v5.8.1, and parts of this will be uploaded to `crates.io`. This should be compatible with the licenses of the included components.

### DIFF
--- a/suitesparse_bindings/suitesparse-src/Cargo.toml
+++ b/suitesparse_bindings/suitesparse-src/Cargo.toml
@@ -9,6 +9,51 @@ license-file = "SuiteSparse/LICENSE.txt"
 readme = "README.md"
 repository = "https://github.com/vbarrielle/sprs"
 description = "Builds the SuiteSparse components"
+exclude = [
+    "SuiteSparse/AMD/",
+    "SuiteSparse/bin/",
+    "SuiteSparse/BTF/",
+    "SuiteSparse/CCOLAMD/",
+    "SuiteSparse/CHOLMOD/",
+    "SuiteSparse/COLAMD/",
+    "SuiteSparse/CSparse/",
+    "SuiteSparse/CSparse_to_CXSparse/",
+    "SuiteSparse/CXSparse/",
+    "SuiteSparse/CXSparse_newfiles/",
+    "SuiteSparse/GPUQREngine/",
+    "SuiteSparse/GraphBLAS/",
+    "SuiteSparse/KLU/",
+    "SuiteSparse/lib/",
+    "SuiteSparse/Makefile",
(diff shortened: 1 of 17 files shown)


suitesparse_bindings/suitesparse-src/Cargo.toml
```diff
@@ -0,0 +1,65 @@
+[package]
+name = "suitesparse-src"
+version = "0.1.0"
```
vbarrielle, line 3: Just a thought: should this version match the suitesparse version?

vbarrielle (approved):
This is very neat! As I mentionned in a comment, do you think the `suitesparse-src` crate should have its version match the version of SuiteSparse compiled?

vbarrielle:
Another question: I guess you'll want to be the one uploading the `suitesparse-src` crate to crates.io?

suitesparse_bindings/suitesparse-src/Cargo.toml
mulimoen: I find it's easier to leave these two versions separate, and rather have an easy way to find the linked version, pushed a new commit for this. This way it is easier to introduce breaking changes, which we will probably have to do as we iterate on this design

mulimoen:
> Another question: I guess you'll want to be the one uploading the suitesparse-src crate to crates.io?

I don't really mind. You may add yourself as an author and publish. Might be easier if you are the owner of all sprs related material.

mulimoen:
As far as I can tell these are not breaking changes for the CAMD stuff, but LDL requires a version bump since I've removed some version constants which were probably incorrect on some platforms and libraries.

suitesparse_bindings/suitesparse-src/Cargo.toml
vbarrielle: Ok that's a good rationale.

vbarrielle:
> I don't really mind. You may add yourself as an author and publish. Might be easier if you are the owner of all sprs related material.

Ok I'll do it that way then.

> As far as I can tell these are not breaking changes for the CAMD stuff, but LDL requires a version bump since I've removed some version constants which were probably incorrect on some platforms and libraries.

Noted, I'll push the updated version soon.

vbarrielle:
Thanks for all the good work!

PR #236, opened 2020-10-02
Checked constructors for CsVec and CsMat

mulimoen:
This is a first step in disallowing invalid structures from being created, to uphold the guarantees of sortedness of the structures.
This limits the surface where non-sortedness could appear by calling `new_trusted`. We could add debug assertations to `new_trusted` if we are paranoid.

Methods to test if invariants are upheld has been moved to the `utils` module. They could be useful for library users, and could be made public.

### DIFF
--- a/src/sparse.rs
+++ b/src/sparse.rs
@@ -1,4 +1,5 @@
 use crate::array_backend::Array2;
+use crate::errors::SprsError;
 use crate::indexing::SpIndex;
 use std::ops::Deref;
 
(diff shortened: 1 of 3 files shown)


vbarrielle (approved):
Looks good, this indeed makes the code clearer. If I'm not mistaken this has no effect on the public facing API, so not a breaking change right?

mulimoen:
This should not be a breaking change.

vbarrielle:
Perfect then.

PR #238, opened 2020-10-04
Derive Hash on sparse storage containers

maboesanman:
It would be convenient for the sparse data structures in sprs to implement `Hash`, and deriving it should be fine.

This PR also adds `Eq` to `CsVecBase`.

### DIFF
--- a/src/sparse.rs
+++ b/src/sparse.rs
@@ -78,7 +78,8 @@ pub use self::csmat::CompressedStorage;
 /// [`vstack`]: fn.vstack.html
 /// [`hstack`]: fn.hstack.html
 /// [`bmat`]: fn.bmat.html
-#[derive(Eq, PartialEq, Debug, Copy, Clone)]
+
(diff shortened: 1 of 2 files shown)


vbarrielle (approved):
Looks fine, though I'm quite curious, as I would not have thought it necessary. Do you store sparse containers in a hash map, or do you want `Hash` for another reason?

maboesanman:
I thought I needed this, but soon realized that I wouldn't be able to because I'm using vectors of floats, which don't implement hash...

The idea was to store my sparse vectors in a hash map because the calculation I am doing is likely to perform redundant calculations, so I look up the vector in a hashmap and skip the calculation if it's already present.

I ended up hashing the indices array because it was easier and a better balance of performance and optimization.

Still, the collections being hashable seems useful, even if it didn't pan out in my case

vbarrielle:
I see. Thanks for the info. I also think being hashable for types other than floats can be useful.

PR #241, opened 2020-10-16
Add scalar multiplication for primitive types

mulimoen:
`u8`/`i8`/`u16`/`i16` was not handled with the macro, and was not consistent with the other primitives


vbarrielle:
Good catch, thanks!

PR #242, opened 2020-10-21
Fix shape-bug for CSC in middle_outer_views

kolstae:
Create the correct shape of the view returned from `middle_outer_views` when `self` is a `CSC`


vbarrielle:
Hello and thanks for the fix. This looks good, but could you add a test showing a case that used to fail and is now fixed by these changes?

kolstae:
Absolutely

vbarrielle (approved):

PR #243, opened 2020-10-21
Fix nnz for partial views + add test to expose

kolstae:
There's probably a possibility to have separate nnz implementations, but my rust knowledge is not up to the task. This might not be a hot method, so probably not a big problem.

### DIFF
--- a/src/sparse/csmat.rs
+++ b/src/sparse/csmat.rs
@@ -929,6 +929,7 @@ where
     /// algorithms, which are often linear in the number of non-zeros.
     pub fn nnz(&self) -> usize {
         self.indptr.last().unwrap().index_unchecked()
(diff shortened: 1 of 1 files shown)


vbarrielle:
Thanks for the fix, though it makes me realize there are other places in the code that do not take this into account, for instance https://github.com/vbarrielle/sprs/blob/master/src/sparse.rs#L279. It's also possible there are issues in lots of other functions with these partial views.

I'm going to merge this for now, and I'm going to investigate possible other issues.

> There's probably a possibility to have separate nnz implementations, but my rust knowledge is not up to the task.

I don't think it's possible, and I don't think we want to have separate implementations. I think your fix is more general and could help for other edge cases.

PR #246, opened 2020-11-14
Serde deserialization for IndPtr

mulimoen:
This does not uphold the invariants of `IndPtr`, but this would be easier to introduce in #237

Flattening `IndPtr` gives us backwards compatibility.


vbarrielle:
Thanks for this fix! It's nice that you managed to keep the serialization backwards compatible!

PR #247, opened 2020-11-14
add a PartialEq for indptr and slices

mulimoen:
Advantage is in comparing a slice and an `IndPtr`, without the drawbacks of a `Deref` on `IndPtr` which would ignore the properness


vbarrielle:
Nice, thanks!

PR #248, opened 2020-11-22
use same version of dependencies

mulimoen:
This ensures all dependencies are the same version

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ approx = { version = "0.3.2", optional = true }
 
 [dev-dependencies]
 bencher = "0.1.0"
-tempdir = "0.3.5"
+tempfile = "3.1.0"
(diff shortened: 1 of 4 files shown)


vbarrielle:
Good idea, with the upcoming 0.10 it's the good moment to introduce changes in dependencies versions. Looks like there's a rusftmt issue though.

vbarrielle:
Looking at the failure, it's an issue I've encountered for #245, and it's fixed there. It's caused by a change in rustmft with the latest rust release. Re-running the tests should work.

vbarrielle:
No I'm wrong, it needs to be rebased. Alternatively, I can merge it like this, and the merge result should pass the rustfmt test.

mulimoen:
I've rebased on master now

vbarrielle:
Perfect, thanks!

PR #249, opened 2020-11-23
Some pedantic clippy lints

mulimoen:
This is just a collection of clippy lints which are not enabled by default, but improves code consistency slightly. The documentation should be a bit prettier now with all `doc_markdown` lints fixed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,7 @@ on:
 env:
     CARGO_TERM_COLOR: always
     RUSTFLAGS: -D warnings
+    CLIPPY_DISABLE_DOCS_LINKS: 1
 
 jobs:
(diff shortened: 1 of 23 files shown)


vbarrielle (approved):
Very nice! I'm really not used to clippy, but the changes it brings are definitely worth having!

I have a question though: it looks like there is a complex command line to run all these clippy checks, but I'd like to be able to run them locally with a simple command. Is it not possible to specify the clippy lints used in eg `Cargo.toml`?

mulimoen:
We can specify all these lints in `lib.rs` of every crate we have, but this will make the code somewhat unwieldy. There is still no way of setting workspace-wide linter options for clippy, `clippy.toml` is only for adjusting linter options, not enabling them

PR #251, opened 2020-11-24
Slice friendly binop

mulimoen:
The `binop` methods was determining whether to transpose or not based on `is_standard_layout`. These checks are replaced with fastest and slowest axis instead, closer matching the semantics wanted. This makes matrix slices cheaper to work with during addition and multiplication as we don't have to transpose a `CSR` into `CSC` unnecessarily

### DIFF
--- a/src/sparse.rs
+++ b/src/sparse.rs
@@ -284,6 +284,7 @@ pub trait SparseMat {
 
 pub(crate) mod utils {
     use super::*;
+    use ndarray::Axis;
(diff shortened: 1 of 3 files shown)


vbarrielle (approved):
Very nice, thanks. Unfortunately it looks like merging #249 created some conflicts, so I cannot merge this right now.

mulimoen:
Rebased and green CI

vbarrielle (approved):

Issue #250, opened 2020-11-24
More ergonomic slicing APIs


vbarrielle:
During #245 I noticed that `middle_outer_slice` was not convenient to use. I'm planning to deprecate it and introduce more ergonomic APIs that take `Range` as input.

vbarrielle:
Fixed by #252.

PR #252, opened 2020-12-07
More ergonomic slicing

vbarrielle:
As mentionned in #250, the slicing of middle matrices along the outer dimension is quite cumbersome to use, which is particularly visible in the matrix product implementation in `smmp.rs`. This PR tries to propose a more ergonomic API.

### DIFF
--- a/src/range.rs
+++ b/src/range.rs
@@ -48,3 +48,23 @@ impl Range for std::ops::RangeFull {
         None
     }
 }
+
+impl Range for std::ops::RangeInclusive<usize> {
+    fn start(&self) -> Option<usize> {
+        Some(*self.start())
+    }
+
+    fn end(&self) -> Option<usize> {
+        Some(*self.end() + 1)
+    }
+}
+
+impl Range for std::ops::RangeToInclusive<usize> {
+    fn start(&self) -> Option<usize> {
+        None
+    }
+
+    fn end(&self) -> Option<usize> {
+        Some(self.end + 1)
+    }
+}
--- a/src/range.rs
+++ b/src/range.rs
@@ -1,0 +1,50 @@
+//! This module abstracts over ranges to allow functions in the crate to take
+//! ranges as input in an ergonomic way. It howvers seals this abstractions to
+//! leave full control in this crate.
+
+/// Abstract over `std::ops::{Range,RangeFrom,RangeTo,RangeFull}`
+pub trait Range {
+    fn start(&self) -> Option<usize>;
+
+    fn end(&self) -> Option<usize>;
+}
+
+impl Range for std::ops::Range<usize> {
+    fn start(&self) -> Option<usize> {
+        Some(self.start)
+    }
+
+    fn end(&self) -> Option<usize> {
+        Some(self.end)
+    }
+}
+
+impl Range for std::ops::RangeFrom<usize> {
+    fn start(&self) -> Option<usize> {
+        Some(self.start)
+    }
+
+    fn end(&self) -> Option<usize> {
+        None
+    }
+}
+
+impl Range for std::ops::RangeTo<usize> {
+    fn start(&self) -> Option<usize> {
+        None
+    }
+
+    fn end(&self) -> Option<usize> {
+        Some(self.end)
+    }
+}
+
+impl Range for std::ops::RangeFull {
+    fn start(&self) -> Option<usize> {
+        None
+    }
+
+    fn end(&self) -> Option<usize> {
+        None
+    }
+}
--- a/src/range.rs
+++ b/src/range.rs
@@ -4,14 +4,14 @@
 
 /// Abstract over `std::ops::{Range,RangeFrom,RangeTo,RangeFull}`
 pub trait Range {
-    fn start(&self) -> Option<usize>;
+    fn start(&self) -> usize;
 
     fn end(&self) -> Option<usize>;
 }
 
 impl Range for std::ops::Range<usize> {
-    fn start(&self) -> Option<usize> {
-        Some(self.start)
+    fn start(&self) -> usize {
+        self.start
     }
 
     fn end(&self) -> Option<usize> {
@@ -20,8 +20,8 @@ impl Range for std::ops::Range<usize> {
 }
 
 impl Range for std::ops::RangeFrom<usize> {
-    fn start(&self) -> Option<usize> {
-        Some(self.start)
+    fn start(&self) -> usize {
+        self.start
     }
 
     fn end(&self) -> Option<usize> {
@@ -30,8 +30,8 @@ impl Range for std::ops::RangeFrom<usize> {
 }
 
 impl Range for std::ops::RangeTo<usize> {
-    fn start(&self) -> Option<usize> {
-        None
+    fn start(&self) -> usize {
+        0
     }
 
     fn end(&self) -> Option<usize> {
@@ -40,8 +40,8 @@ impl Range for std::ops::RangeTo<usize> {
 }
 
 impl Range for std::ops::RangeFull {
-    fn start(&self) -> Option<usize> {
-        None
+    fn start(&self) -> usize {
+        0
     }
 
     fn end(&self) -> Option<usize> {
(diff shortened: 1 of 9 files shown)


vbarrielle:
I'm leaving this as-is for now, as I'd like to take some time to reflect on this new API. @mulimoen, if you want, I'd very much appreciate your feedback on these changes.

src/range.rs
```diff
@@ -0,0 +1,50 @@
+//! This module abstracts over ranges to allow functions in the crate to take
+//! ranges as input in an ergonomic way. It howvers seals this abstractions to
+//! leave full control in this crate.
+
+/// Abstract over `std::ops::{Range,RangeFrom,RangeTo,RangeFull}`
```
mulimoen, line 5: Any reason to leave out `RangeInclusive` and `RangeToInclusive`?

src/sparse/slicing.rs
```diff
@@ -0,0 +1,122 @@
+//! This module implementations to slice a matrix along the desired dimension.
+//! We're using a sealed trait to enable using ranges for an idiomatic API.
+
+use crate::range::Range;
+
+impl<N, I, Iptr> crate::CsMatI<N, I, Iptr>
```
mulimoen, line 6: This duplication can be dropped if we implement all the methods on `CsMatBase`, but we would lose the opportunity to use `.view().slice_outer()` (lifetime problems). `.slice_outer()` would work directly

src/sparse/indptr.rs
```diff
@@ -381,6 +381,19 @@ impl<Iptr: SpIndex> IndPtr<Iptr> {
             *val += Iptr::one();
         }
     }
+
+    /// Slice this indptr to include only the outer dimensions in the range
+    /// `start..end`.
+    pub(crate) fn middle_slice(
```
mulimoen, line 387: This method should be implemented on `IndPtrBase`

src/sparse/slicing.rs
mulimoen: And one would need a mutating slicer, `.slice_outer_mut()`. I'm not sure which approach is preferrable

mulimoen (changes requested):
@vbarrielle I think this API is very nice to have, and makes sense to include. I don't think I would change anything apart from the nitpicks noted here.

~~You should do a rebase on current master, the clippy changes I've made will probably give merge conflicts~~
This was me seeing a single missed type disguise, fixed in #258

src/range.rs
vbarrielle: No, I just forgot about them!

src/sparse/slicing.rs
vbarrielle: Yes this topic is a recurring issue, I've tried both approaches over the codebase, and I still can't settle on one approach. What's sure is that there needs to be a method on `CsMatViewI` that returns the lifetime of the type and not the lifetime of `&self`, otherwise some usages are incovenient or impossible.

In the past I've tried three approaches:

- implement on `CsMatBase`, and have another implementation on `CsMatViewI` with another name (e.g. with the suffix `_rbr` to mean the method "reborrows". The upside is that one is guaranteed to have a working implementation of the method, but the downside is that there is some duplication, and the reborrowing method is less discoverable (which can be accounted for through documentation)
- implement only on `CsMatViewI`, requiring to always call `.view().method()` for other types. It's the simplest approach in terms of implementation, with no duplication. But it suffers from a discoverability issue, greater than the previous one. I like this approach though, as it mirrors the relationship between `Vec<T>` and `&[T]`, but there the `Deref`trait makes the relationship seamless.
- Implement explicitly on all types, as here. The upside is discoverability I guess, but it arguably creates a more complex and duplicated implementation. I also risk missing some instatiations of `CsMatBase`.

That's my current state of mind. Ideally I'd like to take this PR as an opportunity to discuss these alternatives, pick one approach and try to be consistent throughout the crate for 0.10.

src/sparse/indptr.rs
vbarrielle: That's the same debate as the previous point, if I implement it on `IndPtrBase` I need to take another name for the "reborrowing" version. I wish it was possible to "specialize" on this lifetime, or perform compile-time lifetime computations to figure out the correct lifetime...

src/range.rs
vbarrielle: I'm wondering now, should `Range::start` trait return an `usize` instead of an `Option<usize>`, with `RangeTo`, `RangeFull` returning `0` in that case? I'm wondering if there can be a downside. Right now I don't see a case where I'd like to know that the start of the range was not explicitly specified.

src/sparse/slicing.rs
vbarrielle: For the third approach, it may be possible to reduce duplication using macros. Another downside I had not mentionned is bloat (but probably similar to implementing on `CsMatBase` because of monomorphization. Which means the second approach has the advantage of generating tighter code.

src/range.rs
mulimoen: You have don't have negative or positive starting offsets so returning an `usize` should be OK

src/sparse/slicing.rs
mulimoen: The third option will lead to N entries on the left hand side of the [docs of CsMatBase](https://docs.rs/sprs/0.9.2/sprs/struct.CsMatBase.html), with different signatures. It would also exclude other items implementing `Deref<[N]>`, such as `Box<[u8]>`, `Cow<u8>` etc.

The first option might be the most consistent, and will prevent a peppering of `view()` all over the place. I also remember having the same sort of problems with temporary views with `ndarray`, so it shouldn't be surprising when such errors does emerge (well, equally suprising anyway).

The discoverability of the second option would not be too bad if we added a note to each applicable function. (this method requires a view, use `view()`)

mulimoen: Why exactly do we need a `reborrow` version of the methods? The following would work:
```rust
let v = csr.view();
let m = v.foo();
```
Is it to keep `m` after `v` goes out of scope?

mulimoen: Regarding monomorphization: We could always have internal methods on `CsMatView`/`CsMatViewMut` and call this from `CsMatBase`. Would be interesting to see the benefits in both compile time and crate size!

vbarrielle: > Why exactly do we need a `reborrow` version of the methods? The following would work:
> 
> ```rust
> let v = csr.view();
> let m = v.foo();
> ```
> 
> Is it to keep `m` after `v` goes out of scope?

There are some situations, like when implementing iterators, where the variable `v` would go out of scope before returning, but we know the data is still valid. See https://github.com/vbarrielle/sprs/blob/6cc2370f685c674886f958994f1d30123c6b0c42/src/sparse/csmat.rs#L1159 for an example.

vbarrielle: > Regarding monomorphization: We could always have internal methods on `CsMatView`/`CsMatViewMut` and call this from `CsMatBase`. Would be interesting to see the benefits in both compile time and crate size!

That's a good idea!

vbarrielle: With this argument about monomorphization, I think we're now down to two options:

- implement on `CsMatBase` a thin wrapper over the real implementation on `CsMatViewI`
- implement on `CsMatViewI` and add documentation comments for readability.

The third approach was the most complicated.

I guess the first approach is preferable, as it makes for the less surprising user experience. Would you agree?

mulimoen: I agree!

src/range.rs
vbarrielle: `RangeInclusive` and `RangeToInclusive` are now implemented, and `Range::start` returns an usize.

src/sparse/indptr.rs
vbarrielle: It's now implemented on `IndPtrBase` with the demonomorphization trick of https://github.com/vbarrielle/sprs/pull/252#discussion_r540551472

vbarrielle:
@mulimoen the latest changes should address all your remarks. I've also added the policy on reborrowing on the guidelines to have a point of reference

mulimoen (approved):

mulimoen:
Excellent work @vbarrielle!

PR #253, opened 2020-12-18
Add  `diag` method to CsMatBase

matzhaugen:
A method to add fetching the diagonal of a matrix. 

I'm quite new to Rust, so I'm not sure if this useful. It's intended to spark a discussion around the implementation. I really wanted the a `diag_view` method, but I'm not sure how to implement it, so ideas are welcome.

### DIFF
--- a/src/sparse/csmat.rs
+++ b/src/sparse/csmat.rs
@@ -1146,11 +1146,14 @@ where
     {
         let shape = self.shape();
         let smallest_dim: usize = cmp::min(shape.0, shape.1);
-        let mut index_vec = Vec::new();
-        let mut data_vec = Vec::new();
-        let mut optional_index: Option<NnzIndex>;
+        // Assuming most matrices have dense diagonals, it seems prudent
+        // to allocate a bit of memory up front
+        let heuristic = smallest_dim / 2;
+        let mut index_vec = Vec::with_capacity(heuristic);
+        let mut data_vec = Vec::with_capacity(heuristic);
+
         for i in 0..smallest_dim {
-            optional_index = self.nnz_index(i, i);
+            let optional_index = self.nnz_index(i, i);
             if let Some(idx) = optional_index {
                 data_vec.push(self[idx].clone());
                 index_vec.push(I::from_usize(i));
@@ -2814,6 +2817,25 @@ mod test {
             vec![0, 3, 4, 5, 8, 10],
             vec![0, 3, 4, 1, 3, 0, 2, 3, 0, 4],
             vec![1, 3, 1, 2, 1, 3, 1, 1, 1, 1],
+        );
+
+        let diag = mat.diag();
+        let expected = CsVec::new(5, vec![0, 1, 3, 4], vec![1, 2, 1, 1]);
+        assert_eq!(diag, expected);
+    }
+
+    #[test]
+    fn diag_rectangular() {
+        // | 1 0 0 3 1 3|
+        // | 0 2 0 0 0 0|
+        // | 0 0 0 1 0 1|
+        // | 3 0 1 1 0 0|
+        // | 1 0 0 0 1 0|
+        let mat = CsMat::new_csc(
+            (5, 6),
+            vec![0, 3, 4, 5, 8, 10, 12],
+            vec![0, 3, 4, 1, 3, 0, 2, 3, 0, 4, 0, 2],
+            vec![1, 3, 1, 2, 1, 3, 1, 1, 1, 1, 3, 1],
         );
 
         let diag = mat.diag();
--- a/src/sparse/csmat.rs
+++ b/src/sparse/csmat.rs
@@ -1140,7 +1140,7 @@ where
     }
 
     /// Get the diagonal of a sparse matrix
-    pub fn diag(&self) -> CsVecI<N, I> 
+    pub fn diag(&self) -> CsVecI<N, I>
     where
         N: Clone,
     {
@@ -1155,7 +1155,7 @@ where
                 Some(idx) => {
                     data_vec.push(self[idx].clone());
                     index_vec.push(I::from_usize(i));
-                },
+                }
                 None => (),
             }
         }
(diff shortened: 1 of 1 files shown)


vbarrielle:
Hello, and thanks for contributing this feature! I'll try and review it soon.

src/sparse/csmat.rs
```diff
@@ -1139,6 +1139,30 @@ where
...
+    /// Get the diagonal of a sparse matrix
+    pub fn diag(&self) -> CsVecI<N, I>
+    where
+        N: Clone,
+    {
+        let shape = self.shape();
+        let smallest_dim: usize = cmp::min(shape.0, shape.1);
+        let mut index_vec = Vec::new();
+        let mut data_vec = Vec::new();
+        let mut optional_index: Option<NnzIndex>;
```
vbarrielle, line 1151: Looks like `optional_index` does not need to be mutable, and could be declared inside the loop.

src/sparse/csmat.rs
```diff
@@ -1139,6 +1139,30 @@ where
...
 
+    /// Get the diagonal of a sparse matrix
+    pub fn diag(&self) -> CsVecI<N, I>
+    where
+        N: Clone,
+    {
+        let shape = self.shape();
+        let smallest_dim: usize = cmp::min(shape.0, shape.1);
+        let mut index_vec = Vec::new();
+        let mut data_vec = Vec::new();
```
vbarrielle, line 1150: I think there could be a heuristic to create `index_vec` and `data_vec` with `Vec::with_capacity`, to minimize reallocations? It's frequent to have sparse matrices with lots of elements on the diagonal. So maybe `let capacity = smallest_dim / 2;` ? What do you think?

src/sparse/csmat.rs
```diff
@@ -2778,6 +2802,25 @@ mod test {
...
+        let mat = CsMat::new_csc(
+            (5, 5),
+            vec![0, 3, 4, 5, 8, 10],
+            vec![0, 3, 4, 1, 3, 0, 2, 3, 0, 4],
+            vec![1, 3, 1, 2, 1, 3, 1, 1, 1, 1],
+        );
+
+        let diag = mat.diag();
+        let expected = CsVec::new(5, vec![0, 1, 3, 4], vec![1, 2, 1, 1]);
+        assert_eq!(diag, expected);
```
vbarrielle, line 2821: Could you also test a non-square matrix?

vbarrielle (changes requested):
This looks good. Could you add the test case with a non square matrix? Thanks.

src/sparse/csmat.rs
matzhaugen: That is a great idea. I am even tempted to preallocate the max possible size, `smallest_dim`.  I presume this would be the most performant, and least memory concerned? I'll defer to you to decide as I'm not an expert.

src/sparse/csmat.rs
vbarrielle: It's actually exactly the same performance. You are right that there are cases where it would be more performant to have a mutable variable declared outside of the loop, but here we have a Plain Old Data type, so whichever way it is declared the compiler will simply reserve the required stack storage (allocation of stack storage is per-function).

An example of pattern where there would be a difference is with a type that involves heap allocation such as `Vec`:

```rust
let mut v = Vec::with_capacity(10);
for i in 0..10 {
    v.clear();
    v.extend(0..i);
   println!("sum: {}", v.iter().sum());
}
```

is more performant than

```rust
for i in 0..10 {
    let mut v = Vec::new();
    v.extend(0..i);
   println!("sum: {}", v.iter().sum());
}
```

because the first version is able to re-use the same heap storage over and over again, while the second one reallocates at each step of the loop.

src/sparse/csmat.rs
vbarrielle: It's hard to guess without enough representative tests. I think this is good for now, and if in the future there comes a need to customize this heuristic we can always add another function with an additional parameter. It's nice that you added `shrink_to_fit`, this will probably be important when there are not that many diagonal elements.

src/sparse/csmat.rs
```diff
@@ -2826,6 +2826,25 @@ mod test {
...
+        let mat = CsMat::new_csc(
+            (5, 5),
+            vec![0, 3, 3, 4, 6, 7],
+            vec![0, 3, 4, 3, 0, 2, 0],
+            vec![1, 3, 1, 1, 3, 1, 1],
+        );
+
+        let diag = mat.diag();
+        let expected = CsVec::new(5, vec![0], vec![1]);
+        assert_eq!(diag, expected);
```
vbarrielle, line 2845: This test does not actually test the capacity of the vector. I'm afraid there's currently no way to get the capacity of the vectors backing a `CsVec`.

You can add it if you want, it should be something like

```rust
    /// The capacity of the vectors backing the storage
    pub fn capacity(&self) -> usize {
        // All methods guarantee that `self.indices` and `self.data` have
        // the same capacity
        self.data.capacity()
    }
```

vbarrielle: But you don't have to, we can trust the standard library to correctly shrink to fit.

vbarrielle (approved):

src/sparse/csmat.rs
matzhaugen: That gives me an error. It seems like `capacity` can only be used on vectors and not slices?
Screen Shot 2020-12-18 at 11 15 16 PM

mulimoen: The method needs to be on `CsVecI`, the owned version of `CsVecBase`.

matzhaugen:
Should I bump the package version and add a changelog btw?

vbarrielle:
> Should I bump the package version and add a changelog btw?

Not on this branch, this branch has breaking changes preparing for a 0.10 version. I will backport your changes to the 0.9 branch and issue a minor release.

vbarrielle:
@matzhaugen I've published version 0.9.3 on crates.io

There is now a [v0.9 branch](https://github.com/vbarrielle/sprs/tree/v0.9) that can be used to backport these kind of changes to 0.9 while we continue iterating on 0.10.

src/sparse/csmat.rs
matzhaugen: Ah, I see, thanks for the explenation

matzhaugen:
Sounds good!

PR #258, opened 2020-12-19
type can be expressed as view

mulimoen:
This `CsMatBase` is actually a `CsMatViewI` in disguise


vbarrielle:
Good catch, thanks!

PR #259, opened 2020-12-19
add iterators for the diagonal

mulimoen:
These are iterating counter-parts to `diag`.

The mutable iterator contains `unsafe`, please check if I have missed anything regarding safety for this.

### DIFF
--- a/src/sparse/csmat.rs
+++ b/src/sparse/csmat.rs
@@ -1177,6 +1177,15 @@ where
         }
     }
 
+    /// Iteration over all entries on the diagonal
+    pub fn diag_iter(
+        &self,
+    ) -> impl ExactSizeIterator<Item = Option<&N>>
+           + DoubleEndedIterator<Item = Option<&N>> {
+        let smallest_dim = cmp::min(self.ncols, self.nrows);
+        (0..smallest_dim).map(move |i| self.get_outer_inner(i, i))
+    }
+
     /// Iteration on outer blocks of size `block_size`
     ///
     /// # Panics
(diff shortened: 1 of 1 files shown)


vbarrielle:
This looks really perfect! I've only had a quick glance so I'm 100% sure for the safety of the mutable version yet, but this sounds very reasonable.

src/sparse/csmat.rs
```diff
@@ -1518,6 +1527,32 @@ where
...
+            if let Some(NnzIndex(idx)) = idx {
+                // To obtain multiple mutable references to different
+                // locations in data we must use a pointer and some unsafe.
+                // # Safety
+                // This is safe as
+                // * NnzIndex provides bounds checking
+                // * diagonal entries are never overlapping in memory
+                // * no entries are requested more than once
+                // * nnz_index_outer_inner does not modify or read from entries in self.data
+                Some(unsafe { &mut *data_ptr.add(idx) })
```
vbarrielle, line 1550: Quick thought, I haven't checked yet if it works: the nnz indices of the diagonal elements should be increasing, so maybe it's possible to keep using safe rust only by using `slice::split_at_mut` and remembering the last nnz index to have the correct offset?

Or we could have an iterator that yields mutable outer views, and chaining that we'd be sure we have at most one reference per mutable outer view.

mulimoen: I was thinking of iterating through the outer views, but could not work out the lifetimes, Could be this is one of the places where we need a reborrowing outer view iterator

vbarrielle: Actually I'm not sure it's possible to use the reborrowing pattern with mutable data as it would break the single reference. So it's possible there's no way other than reimplementing the whole logic for each iterator. Your unsafe impl might be the easiest way actually.

mulimoen: `outer_iterator_mut_rbr` seems to be impossible to create using the current implementation of `outer_iterator`. I noticed `outer_iterator` implements `DoubleEndedIterator` but only works forward using repeated `split_at_mut`. I don't think this is valid?

vbarrielle: You're right, I think I assumed I could do a `DoubleEndedIterator` as in the immutable case, but the current implementation cannot fulfill that bound. I'm not even sure it works in the forward case now, I translated the old implementation to the `impl Iterator` one but as this is not tested a lot I may have introduced issues.

vbarrielle: It's probably possible to implement the `DoubleEndedIterator` by keeping around a range that indicates which part of the original data the current range corresponds to.

vbarrielle (commented):
It looks like it's conflicting with the current master, can you rebase?

mulimoen:
@vbarrielle Rebased with green CI

vbarrielle:
All good!  Thanks.

Issue #256, opened 2020-12-19
Serde tests failure


vbarrielle:
I've recently noticed that the test `tests/bincode_ser.rs` was not being executed. I fixed that to make it run in the [fix_serde_test_branch](https://github.com/vbarrielle/sprs/tree/fix_serde_test), but it reveals there's an error at serialization time for `CsMatView`: https://github.com/vbarrielle/sprs/runs/1581558055

I've tried looking into it, but so far I don't get what's wrong, the serialization code is the same than for `CsVec`, where it works... @mulimoen you know that serialization code much better than I do, do you have an idea?

vbarrielle:
So the following diff fixes the issue:

```diff
diff --git a/src/sparse.rs b/src/sparse.rs
index c3fb133..0f86458 100644
--- a/src/sparse.rs
+++ b/src/sparse.rs
@@ -102,7 +102,7 @@ where
     storage: CompressedStorage,
     nrows: usize,
     ncols: usize,
-    #[cfg_attr(feature = "serde", serde(flatten))]
+    //#[cfg_attr(feature = "serde", serde(flatten))]
     indptr: IndPtrBase<Iptr, IptrStorage>,
     indices: IndStorage,
     data: DataStorage,
```

Looks like flattening the `IndPtrBase` is causing troubles, but I don't know why.

vbarrielle:
I've found a similar issue https://github.com/servo/bincode/issues/245

vbarrielle:
And according to https://github.com/servo/bincode/issues/167#issuecomment-653828999 this is not something we can be expecting a fix for.

Looks like we have two alternatives:
- implement Serialize/Deserialize manually
- break backwards compatibility on the serialization format

I'd really prefer the first solution.

vbarrielle:
Something I don't get though: by removing the `flatten` attribute, serialization works, but I would expect the deserialization to fail since there should be one more level of structure. Or maybe that's not exposed with bincode.

vbarrielle:
And indeed, if I test a roundtrip with `serde_json` without the `flatten` attribute, there is an error because what's serialized is no longer what was expected.

Looks like there's only `Serialize` that needs to be implemented manually.

mulimoen:
I think we can cheat a little here and use the `into` container attribute. I'll test this approach

mulimoen:
[...]

PR #260, opened 2020-12-20
Fix serde serialization

vbarrielle:
As noted in #256, there was a failure when serializing with bincode. This PR should fix it by implementing `Serialize` manually. See https://github.com/servo/bincode/issues/245 for the source of the issue.

### DIFF
--- a/src/sparse/serde_traits.rs
+++ b/src/sparse/serde_traits.rs
@@ -1,6 +1,33 @@
 use super::*;
-pub(crate) use serde::{Deserialize, Serialize};
+pub(crate) use serde::ser::SerializeStruct;
+pub(crate) use serde::{Deserialize, Serialize, Serializer};
 use std::convert::TryFrom;
+use std::ops::Deref;
+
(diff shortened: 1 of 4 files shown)


src/sparse/serde_traits.rs
```diff
@@ -1,6 +1,33 @@
...
+    N: Serialize,
+    IptrStorage: Deref<Target = [Iptr]>,
+    IStorage: Deref<Target = [I]>,
+    DStorage: Deref<Target = [N]>,
+{
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut state = serializer.serialize_struct("CsMatBase", 3)?;
```
mulimoen, line 21: Should this be `6` instead of `3`?

vbarrielle: Oops I don't know how that slipped through. It's fixed now.

mulimoen (approved):

PR #262, opened 2020-12-21
add miri workflow

mulimoen:
If we have to add some unsafe code we should ensure we don't do it wildly wrong. I suggest adding `miri` to the CI which could catch some undefined behaviour.

The `io` module can not be tested with `miri` as file operations are not supported


vbarrielle:
That's a good idea indeed, thanks!

Issue #261, opened 2020-12-21
outer_iterator_mut does not implement DoubleEndedIterator correctly


mulimoen:
`outer_iterator_mut` uses a successive application of `split_at_mut`. The iterator is such not double-ended, and will fail when trying to iterate backwards on it. The code for this can be seen at https://github.com/mulimoen/sprs/tree/bug/outer_iterator_mut, see the actions page at https://github.com/mulimoen/sprs/actions/runs/436352486 for the error message.

How to fix: We probably need to sprinkle a bit of unsafe here to be able to move `self.data` into the closure, and make it independent of iteration order.

vbarrielle:
Fixed by #263

PR #263, opened 2020-12-21
Fix outer_iterator_mut DoubleEnded iterator

mulimoen:
Test and fix for #261 

Introduces unsafe code, please review carefully

### DIFF
--- a/tests/outer_iterator.rs
+++ b/tests/outer_iterator.rs
@@ -1,0 +1,37 @@
+#[test]
+fn outer_iterator() {
+    use sprs::{CsMat, CsVec};
+    // | 1 0 0 3 1 |
+    // | 0 2 0 0 0 |
+    // | 0 0 0 1 0 |
+    // | 3 0 1 1 0 |
+    // | 1 0 0 0 1 |
+    let mut mat = CsMat::new_csc(
+        (5, 5),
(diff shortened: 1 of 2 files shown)


tests/outer_iterator.rs
```diff
@@ -0,0 +1,37 @@
...
+    assert_eq!(second.view(), CsVec::new(5, vec![1], vec![2]).view());
+    let fourth = iter.next_back().unwrap();
+    assert_eq!(
+        fourth.view(),
+        CsVec::new(5, vec![0, 2, 3], vec![3, 1, 1]).view()
+    );
+    let third = iter.next_back().unwrap();
+    assert_eq!(third.view(), CsVec::new(5, vec![3], vec![1]).view());
+
+    assert_eq!(iter.next(), None);
```
vbarrielle, line 36: It would be nice to also exercise the mutability in this test, eg modifying some values.

vbarrielle:
That looks good, I'm quite convinced this unsafe is completely ok. It's also way simpler than any safe alternative I try to think of.

tests/outer_iterator.rs
vbarrielle: Though that is already tested in another test.

vbarrielle (approved):

Issue #257, opened 2020-12-19
CsVecBase::to_dense()


matzhaugen:
I was wondering if there is currently a way to get the dense version of a sparse vector, just like the CsMat::to_dense() equivalent. This is to do element-wise multiplication between dense and sparse vectors.

This is for the following use case:

```
result = ndarray::Zip::from(&mut dense_vector.diag_mut())
                .and(&sparse_vector.diag().to_dense())
                .apply(|a, b| *a -= 1.0 / b);
```

Or another similar use case

```
dense_vector.diag_mut() = dense_vector.diag() - 1 / &sparse_vector.diag();
```

Not sure if the pointers are exactly correct here, but hopefully you get the idea.

mulimoen:
The idiomatic way would be to use an iterator:
```rust
for (i, n) in sparse_diag.iter() {
    dense_vector[i] -= 1.0 / n;
}
```
There is no current "densifier" of sparse vectors. If we want this implemented should we prefer to return a `Vec` or `ndarray::Array1`?

matzhaugen:
Thanks for the snippet!

It seems like Array1 would be consistent with the previous implementation for CsMat
https://github.com/vbarrielle/sprs/blob/master/src/sparse/csmat.rs#L1036

vbarrielle:
Yes it should be implemented on `Array1`. The best way to do it if you want to do a PR is to implement it in the `src/sparse/to_dense.rs` file. It should be possible to extract the method you're interested in from `assign_to_dense`.

vbarrielle:
Fixed by #264.

PR #264, opened 2020-12-22
Add CsVecBase::to_dense()

matzhaugen:
Add method `CsVecBase::to_dense()`

### DIFF
--- a/src/sparse/to_dense.rs
+++ b/src/sparse/to_dense.rs
@@ -1,6 +1,6 @@
-use super::CsMatViewI;
+use super::{CsMatViewI, CsVecViewI};
 use crate::indexing::SpIndex;
-use crate::Ix2;
+use crate::{Ix1, Ix2};
 ///! Utilities for sparse-to-dense conversion
 use ndarray::{ArrayViewMut, Axis};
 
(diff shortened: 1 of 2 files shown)


src/sparse/to_dense.rs
```diff
@@ -75,4 +75,39 @@ mod test {
...
+    N: Clone,
+    I: SpIndex,
+{
+    if spvec.dim() != array.len() {
+        panic!("Dimension mismatch");
+    }
+
+    for (ind, val) in spvec.iter() {
+        array[[ind]] = val.clone();
+    }
```
vbarrielle, line 112: Bt convention, implementations are generally defined above the `test` module, which makes it easier to find them. Could you move that up?

vbarrielle (changes requested):
Thanks for this implementation. It's looking good, I'd just like to have the implementation above the `test` module, and near `assign_to_dense`, to be able to easily browse the file. Thanks!

vbarrielle:
Fixes #257.

src/sparse/to_dense.rs
matzhaugen: Sure!

vbarrielle (approved):
All good now, thanks!

PR #268, opened 2021-01-04
Split the error type between structural and linalg errors

### LINKED ISSUE #255: Be able to recover the input type on errors when creating matrices
Right now, when creating types with structure constraints (`IndPtrBase`, `CsMatBase`, `CsVecBase`) from owned data, the owned data is lost if the check fails, which can be incovenient for the caller.

A possible solution is to have `SprsError` be more than an enum, and parameterized on a type, so that it can store the input type in these cases.

I don't know how it will turn out for composability or performance though. Maybe it will be required to box the value, and maybe also to erase its type with `Any` to preserve composability.

This will need a bit of experimentation.

vbarrielle:
Recent work has made the structural constraints of sparse matrices and vectors clearer, making it clear that there can be a single enum discriminating all possible structural errors. Having linalg errors in the same error type was clumsy: structural errors happen when creating a matrix, while linalg errors happen much later, when solving systems or performing other linear algebra operations. Therefore, by splitting the types, it should be easier to react to both kinds of errors.

Since the linalg part of this crate is still small, it's expected that I'm missing possible errors, which is why the error type is a non
exhaustive enum.

Additionaly, I took this refactoring as an opportunity to fix #255, using the same pattern used by the `CsMatBase` and `CsVecBase` constructors to return the storage in case of an error.

@mulimoen I'm very interested in your remarks on the design of these new error types.

### DIFF
--- a/sprs-ldl/src/lib.rs
+++ b/sprs-ldl/src/lib.rs
@@ -56,7 +56,7 @@ use std::ops::IndexMut;
 
 use num_traits::Num;
 
-use sprs::errors::SprsError;
+use sprs::errors::{LinalgError, SingularMatrixInfo};
 use sprs::indexing::SpIndex;
 use sprs::linalg;
 use sprs::stack::DStack;
@@ -189,7 +189,7 @@ impl Ldl {
     pub fn numeric<N, I>(
         self,
         mat: CsMatViewI<N, I>,
-    ) -> Result<LdlNumeric<N, I>, SprsError>
+    ) -> Result<LdlNumeric<N, I>, LinalgError>
     where
         I: SpIndex,
         N: Copy + Num + PartialOrd,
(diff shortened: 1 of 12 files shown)


mulimoen:
Looking at this on my phone it looks great! Will have access to my computer later this week for a proper look.

Would it be an idea to have some error-type which is an enum (`SprsError`) of these two for end-users? Would make migration very easy

vbarrielle:
> Would it be an idea to have some error-type which is an enum (`SprsError`) of these two for end-users? Would make migration very easy

I'm not sure how it would help migration. Constructors will return a `StructureError`, so I don't see how the user would get an enum (`SprsError`) of `LinalgError` and `StructureError`.

Something that could be done to ease migration, would be to have a type alias `pub type SprsError = StructureError`, which would mean constructors would not need much migration, but linalg functions would. But since the variants have changed, it's not a guarantee there would be no breakage.

vbarrielle:
Ok I get it now, looks nice indeed, I'll add it. Thanks!

vbarrielle:
@mulimoen the `SprsError` type you suggested is now present.

mulimoen:
I had another look, and the PR seems ready to merge. Excellent work @vbarrielle, will make future error handling a breeze

PR #269, opened 2021-01-13
Triplet matrix conversion to compressed storage with given Iptr

kolstae:
It would be good to convert triplet matrix to a cs with chosen `Iptr` type.
There might be a way to let `Iptr = I` as default?

### DIFF
--- a/src/io.rs
+++ b/src/io.rs
@@ -508,7 +508,7 @@ mod test {
     fn read_write_read_matrix_market_via_csc() {
         let path = "data/matrix_market/simple.mm";
         let mat = read_matrix_market::<f64, usize, _>(path).unwrap();
-        let csc = mat.to_csc();
+        let csc: CsMat<_> = mat.to_csc();
         let tmp_dir = tempdir().unwrap();
         let save_path = tmp_dir.path().join("simple_csc.mm");
         write_matrix_market(&save_path, &csc).unwrap();
--- a/src/sparse/triplet_iter.rs
+++ b/src/sparse/triplet_iter.rs
@@ -108,7 +108,7 @@ where
     DI: Clone + Iterator<Item = &'a N>,
 {
     /// Consume `TriMatIter` and produce a CSC matrix
-    pub fn into_csc(self) -> CsMatI<N, I>
+    pub fn into_csc<Iptr: SpIndex>(self) -> CsMatI<N, I, Iptr>
     where
         N: Num,
     {
@@ -116,7 +116,7 @@ where
     }
 
     /// Consume `TriMatIter` and produce a CSR matrix
-    pub fn into_csr(self) -> CsMatI<N, I>
+    pub fn into_csr<Iptr: SpIndex>(self) -> CsMatI<N, I, Iptr>
     where
         N: Num,
     {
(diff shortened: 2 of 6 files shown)


kolstae:
Not sure how to fix CI / test (nightly). Doesn't seem to be related to my change

mulimoen:
CI can be fixed by removing `-Zpackage-features` from line 128 in `ci.yml`. This was a check for consistency of build-flags, but has now been stabilised

kolstae:
> CI can be fixed by removing `-Zpackage-features` from line 128 in `ci.yml`. This was a check for consistency of feature-flags, but has now been stabilised

Should I do this in this PR though?

src/io.rs
```diff
@@ -508,7 +508,7 @@ mod test {
     fn read_write_read_matrix_market_via_csc() {
         let path = "data/matrix_market/simple.mm";
         let mat = read_matrix_market::<f64, usize, _>(path).unwrap();
-        let csc = mat.to_csc();
+        let csc = mat.to_csc::<usize>();
```
vbarrielle, line 511: An alternative to the turbofish syntax would be to constrain the type of `csc`, something like:

```rust
let csc: crate::CsMat = mat.to_csc();
```

For some reason I tend to prefer this syntax, but it's only a matter of taste.

src/sparse/triplet_iter.rs
```diff
@@ -108,23 +108,26 @@ where
     DI: Clone + Iterator<Item = &'a N>,
 {
     /// Consume `TriMatIter` and produce a CSC matrix
-    pub fn into_csc(self) -> CsMatI<N, I>
+    pub fn into_csc<Iptr: SpIndex>(self) -> CsMatI<N, I, Iptr>
```
vbarrielle, line 111: > There might be a way to let Iptr = I as default?

Unfortunately I don't think it's enabled on functions, see https://github.com/rust-lang/rust/issues/36887 (looks like there is some debate though). For the time being I think we can accept the occasional need for type annotations, especially since the type inference might actually help in functions larger than the tests.

vbarrielle (commented):
Thanks for this PR, this was indeed lacking. It's also perfectly timed, since that breaking change can be bundled with the others in 0.10.

> > CI can be fixed by removing `-Zpackage-features` from line 128 in `ci.yml`. This was a check for consistency of feature-flags, but has now been stabilised
> 
> Should I do this in this PR though?

I'm ok with this CI fix done in this PR, as long as it's in a separate commit.

Otherwise it's all looking good.

src/io.rs
kolstae: If you would like me to switch to that syntax just let me know.

kolstae: Never mind, I'm switching

mulimoen:
We can avoid the extra type by changing `TripletMatBase`  to have separate types for row and columns with a default type, as done with `CsMatBase`. This is a ton more work with little benefit, tri matrices are only intended for construction as of now.

I'm happy with type annotations, they should be elided by type coercion in most cased as mentioned by vbarrielle

kolstae:
I'm sorry but I'm not able make cargo build `sprs-ldl` with `sprs_suitesparse_ldl` and `sprs_suitesparse_camd` enabled. Could you help me?

mulimoen:
Did nigthly break the `-p` option? Seems you have to specify package now, use `--features sprs-ldl/sprs_suitesparse_ldl`

vbarrielle:
All right, thanks @kolstae and @mulimoen for figuring out the CI fix. Specifying features of dependencies is clearly a pain point in rust...

Issue #93, opened 2016-10-10
Sparse mat / dense vector product needs cleanup


vbarrielle:
Right now this product is either available as a special case of the sparse/dense matrix multiplication, or as a free function that only takes slices as input.

We need to rewrite the free function to take ndarray vectors as input, and write the operator for sparse matrix/dense vector in term of this function.

toastronics:
Recap, to make sure I understand what's being proposed here -

The change is related to definitions in `src/sparse/prod.rs` and `src/sparse/csmat.rs`.

"Special case of the sparse/dense matrix multiplication":

  - `impl Mul<&ArrayBase<Ix2>> for &CsMatBase`;
  - the free function it calls, `{csr, csc}_mulacc_dense_rowmaj(lhs, rhs, out)`, where `lhs` is an N x M matrix;
  - in the special case where `rhs` and `out` are M x 1 matrices.

"Free function that only takes slices as input": `mul_acc_mat_vec_{csr, csc}`

Proposed change:

  - modify the interface of `mul_acc_mat_vec_{csr,csc}(mat, in_vec, res_vec)` such that `in_vec` and `res_vec` are `ArrayBase<Ix1>` instead of slices;
  - add an `impl Mul<&ArrayBase<Ix1>> for &CsMatBase` using the modified `mul_acc_mat_vec_{csr,csc}`.

Is this understanding correct?

Another possible option, avoiding an API change, would be to keep `mul_acc_mat_vec_{csr,csc}` as-is and take advantage of the `ndarray::ArrayView` method `into_slice(&self) -> Option<&'a [N]>`. However, this method returns `None` if the `ArrayView` is not "contiguous and in standard order"; an `impl Mul<&ArrayBase<Ix1>> for &CsMatBase` which made use of `into_slice()` would not be as general as the type signature implies. Due to this restriction, the proposed change in the signature of `mul_acc_mat_vec_{csr,csc}` seems like the better option if this minor break in the API is acceptable; it is simple enough to construct an `ArrayBase<Ix1>` from a slice if required for any existing external callers of `mul_acc_mat_vec_{csr,csc}`.

[...]

PR #272, opened 2021-01-20
Draft: Support generic dense vectors in more APIs

vbarrielle:
Lots of APIs were using slices as their input, particularly when using
output buffers. However, this was not a good fit for linear algebra,
since consumers of the APIs are more likely to be using eg an array from
ndarray.

Here we extend the `DenseVector` trait with a mutable version to be able
to use it on output parameters. Since these are sealed traits we should
be able to add unsafe indexing if necessary without a breaking change.
It should also be possible to support eg nalgebra in the future without
too much trouble.

This should improve the situation discussed in #93, though it's
probably not done yet.

### DIFF
--- a/src/dense_vector.rs
+++ b/src/dense_vector.rs
@@ -15,6 +15,17 @@ pub trait DenseVector<N> {
     fn index(&self, idx: usize) -> &N;
 }
 
+impl<N> DenseVector<N> for [N] {
+    fn dim(&self) -> usize {
+        self.len()
+    }
+
+    #[inline(always)]
+    fn index(&self, idx: usize) -> &N {
+        &self[idx]
+    }
+}
+
 impl<'a, N: 'a> DenseVector<N> for &'a [N] {
     fn dim(&self) -> usize {
         self.len()
@@ -91,6 +102,13 @@ pub trait DenseVectorMut<N>: DenseVector<N> {
     ///
     /// If the index is out of bounds
     fn index_mut(&mut self, idx: usize) -> &mut N;
+}
+
+impl<'a, N: 'a> DenseVectorMut<N> for [N] {
+    #[inline(always)]
+    fn index_mut(&mut self, idx: usize) -> &mut N {
+        &mut self[idx]
+    }
 }
 
 impl<'a, N: 'a> DenseVectorMut<N> for &'a mut [N] {
--- a/src/dense_vector.rs
+++ b/src/dense_vector.rs
@@ -1,0 +1,125 @@
+use crate::Ix1;
+use ndarray::{self, ArrayBase};
+
+/// A trait for types representing dense vectors, useful for expressing
+/// algorithms such as sparse-dense dot product, or linear solves.
+pub trait DenseVector<N> {
+    /// The dimension of the vector
+    fn dim(&self) -> usize;
+
+    /// Random access to an element in the vector.
+    ///
+    /// # Panics
+    ///
+    /// If the index is out of bounds
+    fn index(&self, idx: usize) -> &N;
+}
+
+impl<'a, N: 'a> DenseVector<N> for &'a [N] {
+    fn dim(&self) -> usize {
+        self.len()
+    }
+
+    #[inline(always)]
+    fn index(&self, idx: usize) -> &N {
+        &self[idx]
+    }
+}
+
+impl<'a, N: 'a> DenseVector<N> for &'a mut [N] {
+    fn dim(&self) -> usize {
+        self.len()
+    }
+
+    #[inline(always)]
+    fn index(&self, idx: usize) -> &N {
+        &self[idx]
+    }
+}
+
+impl<N> DenseVector<N> for Vec<N> {
+    fn dim(&self) -> usize {
+        self.len()
+    }
+
+    #[inline(always)]
+    fn index(&self, idx: usize) -> &N {
+        &self[idx]
+    }
+}
+
+impl<'a, N: 'a> DenseVector<N> for &'a Vec<N> {
+    fn dim(&self) -> usize {
+        self.len()
+    }
+
+    #[inline(always)]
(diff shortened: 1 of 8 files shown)


vbarrielle:
This will probably require some iteration. @mulimoen I'd be very interested in your feedback :)

mulimoen:
This looks great! I tried implementing `DenseVector` on `Deref<Target = [N]>` to replace some of the boilerplate implementations, but this seems to clash with `ndarray::ArrayBase`, even though this does not implement `Deref`. It seems rust has some forward protection based on the error message:
```rust
note: upstream crates may add a new impl of trait `std::ops::Deref` for type `ndarray::ArrayBase<_, ndarray::Dim<[usize; 1]>>` in future versions
```

mulimoen:
I was thinking about incorporating the `nshare` crate, but I don't think this is a good idea. Adding support for `nalgebra` in the new traits is easy, and we would know there is no overhead. Outside crates should probably implement `DenseVector` themselves, without going through `nshare`.

src/dense_vector.rs
```diff
@@ -0,0 +1,125 @@
+use crate::Ix1;
+use ndarray::{self, ArrayBase};
+
+/// A trait for types representing dense vectors, useful for
+/// defining a fast sparse-dense dot product.
```
mulimoen, line 5: expressing algorithms such as sparse-dense dot product.

src/dense_vector.rs
```diff
@@ -0,0 +1,125 @@
...
+    /// If the index is out of bounds
+    fn index(&self, idx: usize) -> &N;
+}
+
+impl<'a, N: 'a> DenseVector<N> for &'a [N] {
+    fn dim(&self) -> usize {
+        self.len()
+    }
+
+    #[inline]
```
mulimoen, line 23: I think we should use `#[inline(always)]` on these, they should be zero-cost

vbarrielle:
> This looks great! I tried implementing `DenseVector` on `Deref<Target = [N]>` to replace some of the boilerplate implementations, but this seems to clash with `ndarray::ArrayBase`, even though this does not implement `Deref`.

Yes orphan rules will prevent us from having too broad implementations here, I think we need to implement on a case by case. Fortunately most implementations are trivial.

> I was thinking about incorporating the `nshare` crate, but I don't think this is a good idea. Adding support for `nalgebra` in the new traits is easy, and we would know there is no overhead. Outside crates should probably implement `DenseVector` themselves, without going through `nshare`.

Right now the `DenseVector` trait is not publicly exposed by `sprs`, it is sealed. Which is a nice way for us to control precisely the types we want to support. I do think we want to support `nalgebra` better in the future, though probably with a feature flag to control the support.

I haven't played with `nshare` yet, so I don't know if it's convenient to use.

src/dense_vector.rs
vbarrielle: You're right.

vbarrielle:
With the latest changes, I think I'm leaning towards an interesting implementation, I've been able to write a single product impl allowing to apply a permutation to any dense vector (so a `Vec`, a slice, or an `ndarray::Array`).

I'm starting to think the `DenseVector` trait should be exposed outside the crate, I feel the need for it to make the solve in `sprs-ldl` more generic:

```rust
    //I'd like to use DenseVector here, but I probably need to expose it for public use
    pub fn solve<'a, V>(&self, rhs: &V) -> Vec<N>
    where
        N: 'a + Copy + Num,
        V: Deref<Target = [N]>,
    {
        let mut x = &self.symbolic.perm * &rhs[..];
        let l = self.l();
        ldl_lsolve(&l, &mut x);
        linalg::diag_solve(&self.diag, &mut x);
        ldl_ltsolve(&l, &mut x);
        let pinv = self.symbolic.perm.inv();
        &pinv * &x
    }
```

There's another intersting opportunity with this refactor: if I manage to implement everything that's related to dense vectors using these traits, then I can make the `ndarray` dependency optional (but on by default I guess), and also add an optional dependency to `nalgebra` that would also implement this trait. Which is probably not enough, as I'd need also a trait to represent dense matrices to convert from sparse to dense, but it looks like an interesting future direction.

src/dense_vector.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    /// Random access to an element in the vector.
+    ///
+    /// # Panics
+    ///
+    /// If the index is out of bounds
+    fn index(&self, idx: usize) -> &Self::Scalar;
+
+    /// Create an owned version of this dense vector type, filled with zeros
+    fn zeros(dim: usize) -> Self::Owned;
```
mulimoen, line 22: Are you thinking `zeros` will always be the same size as `self` here?

vbarrielle: There's no `self` here, this one is static, I think I'll probably need it for some implementations (like `to_dense`).

mulimoen: I was testing how the trait would interact with const generics. `Self::Owned` can not be constrained to a generic size `[N; M], and would need to be a `Vec` or equivalent.

mulimoen: Not that this is a blocker in any way

vbarrielle: Right, I was thinking a bit about arrays, but I don't have a plan for them right now.

mulimoen: I doubt you'll want to use a sparse library for arrays anyway, those kind of problems are probably a better fit for dense libraries which could use the extra information more efficiently

vbarrielle:
So I think I've reached a good state, I've been able to use `DenseVector` in all required places. @mulimoen if you're OK with the current state then I'll merge.

mulimoen (approved):
This is a massive boost in usability of this crate! Looks great, I say this is ready to be merged.

vbarrielle:
Thanks for the review @mulimoen !

PR #274, opened 2021-01-21
Revamp constructors

### LINKED ISSUE #273: Non-sorting constructor
As defined in https://github.com/vbarrielle/sprs/issues/121#issuecomment-764796038 and the subsequent discussion.

mulimoen:
This moves and renames the constructors of `CsMatBase`. This makes construction of matrices for non-`Copy` types easier, and unifies constructors.

Modified:
* `new`: Now implemented on `CsMatBase` instead of `CsMatI`. Does not sort indices. Panics on failure
* `new_csc`:  Counterpart to above
* `try_new`: Returns a tuple with all owned data + error on failure
* `try_new_csc`: Counterpart to above

New:
* `new_unchecked`: Unsafe constructor for `CsMatBase`, does not do any data checks.

Removed:
* `new_view`: `new` can now be used instead
* `new_view_raw`: This function is really unsafe, we are opening ourselves to a use-after-free by making a lifetime from nothing. I'd rather push this responsibility away from this crate.

Maybe we could remove `new_vecview_raw`? This is not in use by any of our crates, and suffers from the same problems as `new_view_raw`.

Fixes #273


mulimoen:
On second thought I think it would be ok to have both `new` and `try_new`. I will however change `try_new` so it returns the owned data.

src/sparse/csmat.rs
```diff
@@ -249,6 +328,42 @@ where
             }),
         }
     }
+
+    /// Try create an owned `CSR` matrix from moved data.
```
vbarrielle, line 332: Actually it's also possible to create a `CsMatViewMut` using this constructor I think.

src/sparse/smmp.rs
```diff
@@ -439,8 +439,7 @@ mod test {
 
         let mut c_data = vec![0.; c_indices.len()];
         let mut tmp = [0.; 5];
-        let mut c = crate::CsMatViewMutI::new_trusted_mut_view(
-            crate::CSR,
+        let mut c = crate::CsMatViewMutI::new(
```
vbarrielle, line 442: I'm really glad this change manages to make this work, I didn't like the call to `new_trusted_mut_view` I had to introduce before, it felt unidiomatic.

vbarrielle: So now I'm wondering if that means we can remove `new_trusted_mut_view` ?

vbarrielle (approved):
This looks very good, I really like how it enables a more coherent API. As mentionned in the code comments, I'm wondering if it's an opportunity to remove `new_trusted_mut_view` that I introduced in the previous PR, we probably don't want to stabilize this call if it can be expressed with `new` since this refactor.

If it can be removed let's remove it before merging this PR, otherwise I'll merge it as is.

src/sparse/smmp.rs
mulimoen: Yes, that worked, Moved `new_trusted` to the appropriate traits restrictions and this worked

mulimoen: That was a typo it seems, it should call `new_trusted` so we don't check our known valid matrice. Fixed

src/sparse/csmat.rs
mulimoen: I've removed all mentions of `owned` and made some examples on `new` which shows how one can construct different type (aliases) directly

src/sparse/smmp.rs
vbarrielle: Good catch, I had not noticed this would check the invariants unnecessarily.

vbarrielle (approved):
It's perfect now, thanks!

mulimoen:
@vbarrielle What is your though on `new_vecview_raw`? This can easily be replaced by `new_unchecked` as was done with `sprs-ldl`

vbarrielle:
> @vbarrielle What is your though on `new_vecview_raw`? This can easily be replaced by `new_unchecked` as was done with `sprs-ldl`

I think the more we can use `new_unchecked` or `new_trusted` to replace these calls, the better. If we can have a very coherent API it'll be easier to code against `sprs`.

PR #275, opened 2021-01-25
CsVec constructors

### LINKED ISSUE #273: Non-sorting constructor

mulimoen:
This is the vector analouge to #274 

The new API consists of
* `new`: Constructors which panics on failure
* `try_new`: Returns all buffers and the error
* `new_sorted`: A constructors that may sort the indices and data
* `new_unchecked`: Does zero checking of constraints

Internal:
* `new_trusted`: The internal version of `new_unchecked`

A bunch of the other constructors were removed. These can be expressed in some form of the above. The `unsafe` view constructors are removed in preference to `new_unchecked`, where the caller must ensure valid lifetimes

Fixes #273

### DIFF
--- a/src/sparse/vec.rs
+++ b/src/sparse/vec.rs
@@ -486,6 +486,7 @@ where
     }
 }
 
+/// # Constructor methods
 impl<N, I: SpIndex, DStorage, IStorage> CsVecBase<IStorage, DStorage, N, I>
 where
     DStorage: std::ops::Deref<Target = [N]>,
@@ -555,6 +556,17 @@ where
     /// - if `indices` and `data` lengths differ
     /// - if the vector contains out of bounds indices
     /// - if indices are out of order
+    ///
+    /// # Examples
+    /// ```rust
+    /// # use sprs::*;
+    /// // Creating a sparse owned vector
+    /// let owned = CsVec::new(10, vec![0, 4], vec![-4, 2]);
+    /// // Creating a sparse borrowing vector with `I = u16`
+    /// let borrow = CsVecViewI::new(10, &[0_u16, 4], &[-4, 2]);
+    /// // Creating a general sparse vector with different storage types
+    /// let mixed = CsVecBase::new(10, &[0_u64, 4] as &[_], vec![-4, 2]);
+    /// ```
     pub fn new(n: usize, indices: IStorage, data: DStorage) -> Self {
         Self::try_new(n, indices, data)
             .map_err(|(_, _, e)| e)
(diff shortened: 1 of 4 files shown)


src/sparse/vec.rs
```diff
@@ -486,27 +486,150 @@ where
...
+
+impl<N, I: SpIndex, DStorage, IStorage> CsVecBase<IStorage, DStorage, N, I>
+where
+    DStorage: std::ops::DerefMut<Target = [N]>,
+    IStorage: std::ops::DerefMut<Target = [I]>,
+{
+    /// Creates a sparse vector
+    ///
+    /// Will sort indices and data if necessary
+    pub fn new_sorted(
```
mulimoen, line 628: Maybe it would make more sense naming this 'new_unsorted`?

vbarrielle: I think I prefer `new_sorted`, `new_unsorted` makes it sound like we support unsorted indices in the storage. Or maybe `new_from_unsorted` is the more explicit version?

src/sparse/vec.rs
```diff
@@ -486,27 +486,150 @@ where
...
     pub fn try_new(
         n: usize,
-        indices: Vec<I>,
-        data: Vec<N>,
-    ) -> Result<Self, StructureError>
+        indices: IStorage,
+        data: DStorage,
+    ) -> Result<Self, (IStorage, DStorage, StructureError)> {
+        Self::new_(n, indices, data)
+    }
```
vbarrielle, line 583: Looks like `try_new` is actually exactly `new_`, so maybe we can remove `new_` now.

vbarrielle (approved):
Looks good, let's debate a bit on the naming of `new_sorted` vs `new_unsorted` vs `new_from_unsorted`, and maybe remove `new_`, and it'll be done.

Thanks!

src/sparse/vec.rs
mulimoen: Good idea! Done

src/sparse/vec.rs
mulimoen: `new_from_unsorted` is very explicit in what it tries to do. I'll rename to this

mulimoen: Should we rename the matrix version to also be `new_from_unsorted` to have a coherent API?

vbarrielle: Yes indeed!

mulimoen: Renamed in aedc570

vbarrielle (approved):
All good now, thanks a lot!

Issue #121, opened 2017-10-27
Relax Copy bound to Clone


edibopp:
This issue was mentioned in #118. Thus I felt it would be good to make a separate issue for it, as I do not think it makes sense to handle it as part of that. Big integers or rationals would be an interesting use case, for example.

mulimoen:
@vbarrielle While we are doing some breaking changes: Maybe we could change the constructor methods of `CsMatBase` so `new` does not do any sorting and we add a new method `new_sort` which acts as `new` does today?

vbarrielle:
@mulimoen are you sure you wanted to comment on this issue? It looks like you wanted to comment on #232.

vbarrielle:
But this particular issue could be solved for 0.10.

mulimoen:
It is a big stopper here too, the new constructors would not require elements being `Clone` nor `Copy`

vbarrielle:
I see, what would the behavior of the non-sorting `new` constructor be? Error returning, same as the other structure checks?

mulimoen:
I think error returning is the cleanest, we would need this constructor anyway. We can always rename and add a panicking one if we find this cumbersome.

vbarrielle:
Ok, let's include the `new` and `new_sorted` constructors for 0.10 then. As you say, a panicking constructor can always be added later. I'm goind to add a separate issue as in my mind this one was more about removing the `Copy` bound that's present on most linear algebra functions.

vbarrielle:
Fixed by #276.

PR #276, opened 2021-03-05
Relax copy bound

vbarrielle:
This pull request removes the `Copy` bound from most functions and methods, fixing #121.

While removing the bound, I tried to ensure there were not too many unnecessary calls to `Clone`, with the idea that this should enable storing more complex data as our scalar type (such as big ints, or heap-allocated matrices to have sparse by blocks matrices) while remaining performant.

The main way the unnecessary clones have been removed is through the introduction of the `MulAcc` trait, which represents the fundamental operation for matrix products. Since it takes its arguments by reference, it should be possible to implement this trait without unnecessary allocations for heab-based scalar types. For convenience, this trait is automatically implemented for all `Copy` types that also implement `num_traits::MulAdd`. This ensures all common scalar types work, and should also help the compiler using the fused-multiply-add instruction when available.

Another pattern that was useful to avoid unnecessary clones was adding bounds such as `for <'r> &'r N: std::ops::Add<&'r N, Output=N>`. These bounds exist for all classic scalar types. However, it can be hard to use them in traits, as evidenced by https://github.com/rust-lang/rust/issues/82779. Because of this issue, implementations of `std::ops::{Add,Sub}` for `CsVecBase` were forced to use one `Clone`.

I have not relaxed the `Copy` bound in `sprs-ldl` yet, I think I would need to add a trait such as `MulSubAcc` similar to `MulAcc` that could be used in solvers. Otherwise I would need to sprinkle the code with calls to `clone()` everywhere.

@mulimoen as usual I'd be very interested in your feedback.

### DIFF
--- a/src/sparse/kronecker.rs
+++ b/src/sparse/kronecker.rs
@@ -11,14 +11,13 @@ use crate::sparse::prelude::*;
 ///
 /// * if indices are out of bounds for its type
 #[must_use]
-pub fn kronecker_product<
-    N: num_traits::Num + Copy + Default,
-    I: SpIndex,
-    Iptr: SpIndex,
->(
+pub fn kronecker_product<N: Clone + Default, I: SpIndex, Iptr: SpIndex>(
     mut a: CsMatViewI<N, I, Iptr>,
     mut b: CsMatViewI<N, I, Iptr>,
-) -> CsMatI<N, I, Iptr> {
+) -> CsMatI<N, I, Iptr>
+where
+    for<'r> &'r N: std::ops::Mul<&'r N, Output = N>,
+{
     use crate::CompressedStorage::CSR;
     if a.storage() == b.storage() {
         let was_csc = a.is_csc();
@@ -38,8 +37,8 @@ pub fn kronecker_product<
         indptr.push(element_count);
         for a in a.outer_iterator() {
             for b in b.outer_iterator() {
-                for (ai, &a) in a.iter() {
-                    for (bi, &b) in b.iter() {
+                for (ai, a) in a.iter() {
+                    for (bi, b) in b.iter() {
                         indices.push(I::from(ai * b_shape.1 + bi).unwrap());
                         element_count += Iptr::one();
                         values.push(a * b);
--- a/src/sparse/kronecker.rs
+++ b/src/sparse/kronecker.rs
@@ -1,5 +1,41 @@
 use crate::indexing::SpIndex;
 use crate::sparse::prelude::*;
+use num_complex::{Complex32, Complex64};
+
+/// Trait for types that are valid to compute a Kronecker product.
+/// This includes all classic scalar types, but can be extended to matrices
+/// as well, which will enable computing the kronecker product of sparse
+/// by block matrices.
+pub trait Kronecker<B = Self> {
+    type Output;
+    fn kron(&self, b: &B) -> <Self as Kronecker<B>>::Output;
+}
+
+macro_rules! scalar_kronecker {
+    ($scalar: ident) => {
+        impl Kronecker for $scalar {
+            type Output = Self;
+            fn kron(&self, b: &Self) -> Self::Output {
+                self * b
+            }
+        }
+    };
+}
+
+scalar_kronecker!(u8);
+scalar_kronecker!(i8);
+scalar_kronecker!(u16);
+scalar_kronecker!(i16);
+scalar_kronecker!(u32);
+scalar_kronecker!(i32);
+scalar_kronecker!(u64);
+scalar_kronecker!(i64);
+scalar_kronecker!(isize);
+scalar_kronecker!(usize);
+scalar_kronecker!(f32);
+scalar_kronecker!(f64);
+scalar_kronecker!(Complex32);
+scalar_kronecker!(Complex64);
 
 /// Compute the Kronecker product between two matrices
 ///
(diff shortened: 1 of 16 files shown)


mulimoen (changes requested):
This looks great. @vbarrielle! I only see a problem with the `Kronecker` product, which must be called recursively.

Here is a little snippet you may add to show what this crate is now capable of:
```rust
use num_traits::Zero;
use sprs::*;

#[derive(Clone, Default, Debug, PartialEq)]
struct Mat {
    data: [[i32; 2]; 2],
}

impl Mat {
    fn new(data: [[i32; 2]; 2]) -> Self {
        Self { data }
    }
}

impl Zero for Mat {
    fn zero() -> Self {
        Self {
            data: [[0_i32; 2]; 2],
        }
    }
    fn is_zero(&self) -> bool {
        self == &Self::zero()
    }
}

impl std::ops::Add for Mat {
    type Output = Self;
    fn add(self, other: Self) -> Self {
        let mut out = Self::Output::zero();
        for i in 0..2 {
            for j in 0..2 {
                out.data[i][j] = self.data[i][j] + other.data[i][j]
            }
        }
        out
    }
}

impl std::ops::Mul for &Mat {
    type Output = Mat;
    fn mul(self, other: Self) -> Self::Output {
        let mut out = Self::Output::zero();
        for i in 0..2 {
            for j in 0..2 {
                for k in 0..2 {
                    out.data[i][j] += self.data[i][k] * other.data[k][j];
                }
            }
        }
        out
    }
}

impl MulAcc for Mat {
    fn mul_acc(&mut self, a: &Self, b: &Self) {
        let ab = a * b;
        for i in 0..2 {
            for j in 0..2 {
                self.data[i][j] += ab.data[i][j];
            }
        }
    }
}

#[test]
/// Performs a multiplication of a sparse block-compressed matrix
///
/// Doing this through sparse multiplication gives 2 calls to mul_acc
/// instead of the 8 expected from a dense matrix multiplication
fn matrix_multiply() {
    let mat1 = Mat::new([[1, 2], [3, 4]]);
    let mat2 = Mat::new([[0, -3], [-2, -7]]);
    assert_eq!(&mat1 * &mat1, Mat::new([[7, 10], [15, 22]]));

    //  0  0  1  2
    //  0  0  3  4
    //  1  2  0 -3
    //  3  4 -2 -7
    let smat1 = sprs::CsMat::new(
        (2, 2),
        vec![0, 1, 3],
        vec![1, 0, 1],
        vec![mat1.clone(), mat1, mat2],
    );

    let mat1 = Mat::new([[2, 0], [7, -4]]);
    let mat2 = Mat::new([[0, -99], [9, -7]]);

    //  2  0  0  -99
    //  7 -4  9   -7
    //  0  0  0    0
    //  0  0  0    0
    let smat2 =
        sprs::CsMat::new((2, 2), vec![0, 2, 2], vec![0, 1], vec![mat1, mat2]);

    //  0   0  0    0
    //  0   0  0    0
    // 16  -8 18 -113
    // 34 -16 36 -325
    let smat3 = &smat1 * &smat2;
    assert_eq!(smat3.indptr().raw_storage(), &[0, 0, 2]);
    assert_eq!(smat3.indices(), &[0, 1]);
    let data = smat3.data();
    assert_eq!(data.len(), 2);
    assert_eq!(data[0], Mat::new([[16, -8], [34, -16]]));
    assert_eq!(data[1], Mat::new([[18, -113], [36, -325]]));
}
```

src/sparse/kronecker.rs
```diff
@@ -11,14 +11,13 @@ use crate::sparse::prelude::*;
 ///
 /// * if indices are out of bounds for its type
 #[must_use]
-pub fn kronecker_product<
-    N: num_traits::Num + Copy + Default,
-    I: SpIndex,
-    Iptr: SpIndex,
->(
+pub fn kronecker_product<N: Clone + Default, I: SpIndex, Iptr: SpIndex>(
```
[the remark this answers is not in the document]
vbarrielle: Good idea. It's implemented now.

vbarrielle:
Thanks for the feeback @mulimoen ! I've added a `Kronecker` trait, and included your sparse matrices per block as a test. With a single change from your implementation: I inlined the product implementation in the `MulAcc` impl, to demonstrate that it can be used to remove temporaries.

src/sparse/kronecker.rs
```diff
@@ -1,6 +1,25 @@
...
+impl<N> Kronecker<N> for N
+where
+    for<'r> &'r N: std::ops::Mul<&'r N, Output = N>,
+{
+    type Output = N;
+    fn kron(&self, b: &N) -> <Self as Kronecker<N>>::Output {
+        self * b
+    }
+}
+
```
mulimoen, line 22: I'm afraid this blanket implementation will prevent a block-sparse kronecker product. We should instead use something like the following macro:
```rust
macro_rules! scalar_kronecker {
    ($scalar: ident) => {
        impl Kronecker for $scalar {
            type Output = Self;
            fn kron(&self, b: &Self) -> Self::Output {
                self * b
            }
        }
    };
}

scalar_kronecker!(u8);
scalar_kronecker!(i8);
scalar_kronecker!(u16);
scalar_kronecker!(i16);
scalar_kronecker!(u32);
scalar_kronecker!(i32);
scalar_kronecker!(u64);
scalar_kronecker!(i64);
scalar_kronecker!(isize);
scalar_kronecker!(usize);
scalar_kronecker!(f32);
scalar_kronecker!(f64);
```

mulimoen: And the signature of `kronecker_product` should be
```rust
pub fn kronecker_product<
    Nin: Clone + Default,
    Nout: Clone + Default,
    I: SpIndex,
    Iptr: SpIndex,
>(
    mut a: CsMatViewI<Nin, I, Iptr>,
    mut b: CsMatViewI<Nin, I, Iptr>,
) -> CsMatI<Nout, I, Iptr>
where
    Nin: Kronecker<Output = Nout>,
{
```

vbarrielle: Quite right, I'll change that.

vbarrielle: Should be better now.

mulimoen: Looks perfect!

mulimoen (approved):

vbarrielle:
Thanks for the review @mulimoen!

PR #277, opened 2021-03-17
Version 0.10

vbarrielle:
I think the round of refactoring for 0.10 is complete now, I can't think of some more breaking changes we might want to add. So I think I'm going to publish 0.10 soon. @mulimoen do you see something I could have forgotten?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 
 name = "sprs"
 description = "A sparse matrix library"
-version = "0.9.2"
+version = "0.10.0"
 authors = ["Vincent Barrielle <vincent.barrielle@m4x.org>"]
 edition = "2018"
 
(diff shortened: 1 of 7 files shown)


vbarrielle:
CI failures are expected, I have not pushed `sprs` 0.10 to crates.io, so the dependent crates cannot find it...

mulimoen:
> CI failures are expected, I have not pushed sprs 0.10 to crates.io, so the dependent crates cannot find it...

I think you were just missing a couple of bumps. Let's see if we got them all now

vbarrielle:
Good catch!

mulimoen:
I can't think of any additional breaking changes we need going forwards. I say we publish!

vbarrielle:
It's published! :tada:

PR #279, opened 2021-04-08
Bump ndarray

mulimoen:
`ndarray` had a version bump, but we are still compatible with all versions. This is not a breaking change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,7 +72,7 @@ jobs:
         include:
         - build: msrv
           os: ubuntu-18.04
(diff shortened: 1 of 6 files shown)


mulimoen:
While it would in theory work the current MRSV, the newest `ndarray` version requires `1.49`. This makes the `msrv` CI job a bit tricky, as we can not require `ndarray=0.14` for just this run. I'm leaving this open, and we'll merge it when we gather more breaking changes.

mulimoen:
`MulAcc` is a breaking change, so it is no harm in merging this. Thanks for the reminder @SilvanCodes!

Issue #283, opened 2021-04-21
allow fully parameterized seperate types for matrix elements,vector elements,their products and accumulators


experiment9123:
for AI mixed precision matrices are increasingly common - for example 8bit weights, relying on 32bit accumulation

one way to support this might be this kind of signature:

```
fn matrix_mul(  Matrix<A>,  Vector<B>)-> Vector<Accumulator>
  where   A:Mul<B, Output=Product>
              Product:Add<Output=Accumulator>
             Accumulator:AddAssign<Product> + Default
```

it may however be even more intuitive to rely on the 'accumulator' being reverse infered , and rely on it supporing "+=" for products.
this would support all eventualities, e.g. 16bit vector values, 8bit matrix weights -> 24 bit products(in i32), but so many components that 64bit accumulators are needed.

I wasn't sure if the library already supports this but I see the "MulAcc" trait which I presume is used for the evaluation. Perhaps the type relationships here could be retrofitted to an extention of this trait

This would also allow for *dimension checked types*, e.g. physical units for the A,B, Outputs being seperate. However merely considering that will miss the potential for product / accumulator precision

Furthermore putting non numeric types into these would allow using the sparse matrix ops as a way of performing graph operations (for example the "Accumulator" could be a collection that takes information from edge links described by a sparse matrix)

mulimoen:
The `MulAcc` trait was added to avoid a `Copy`, and not for accumulating into another type, but this would be a very interesting extension. As far as I can tell, there is no reason why this has not been tried apart from a lack of time and interest.

If you wish to implement this, the easiest way is to copy `mul_acc_mat_vec_csr` in `src/spase/prod.rs`, and tweak the bound on `N` to instead rely on a combination of the `num_traits` traits. I see `uom` is using `num_traits`, so the physical units should be covered by such an approach.

experiment9123:
https://docs.rs/sprs/0.10.0/sprs/trait.MulAcc.html

now examining more closely I see that you really do rely on *Fused* multiply add. it might take a bit more thought for me how this will work.
I'm aware that *fused* multiply add has different precision and hardware support in many machines, so is important for people to be able to express explicitely

I think it might be possible even in my experiment and still allowing everything as generic as I have in mind.. but I should confirm this before trying anything here

mulimoen:
> `T: MulAdd<VecElem,Acc,Output=Acc>`

[...]

PR #284, opened 2021-04-24
fully generic MulAcc<A,B> - groundwork for Fully Generic Matrix/Vector element, Product and Accumulator Types - 

experiment9123:
Partial implementation of #283

### DIFF
--- a/src/mul_acc.rs
+++ b/src/mul_acc.rs
@@ -12,16 +12,19 @@
 /// to provide the most performant implementation. For instance, we could have
 /// a default implementation for numeric types that are `Clone`, but it would
 /// make possibly unnecessary copies.
-pub trait MulAcc {
+pub trait MulAcc<A = Self, B = A> {
     /// Multiply and accumulate in this variable, formally `*self += a * b`.
-    fn mul_acc(&mut self, a: &Self, b: &Self);
+    fn mul_acc(&mut self, a: &A, b: &B);
 }
 
-impl<N> MulAcc for N
+/// Default for types which supports `mul_add`
+impl<N, A, B> MulAcc<A, B> for N
 where
-    N: Copy + num_traits::MulAdd<Output = N>,
+    N: Copy,
+    B: Copy,
+    A: num_traits::MulAdd<B, N, Output = N> + Copy,
 {
-    fn mul_acc(&mut self, a: &Self, b: &Self) {
+    fn mul_acc(&mut self, a: &A, b: &B) {
         *self = a.mul_add(*b, *self);
     }
 }
@@ -37,5 +40,23 @@ mod tests {
         let c = 3.;
         a.mul_acc(&b, &c);
         assert_eq!(a, 7.);
+    }
+
+    #[derive(Debug, Copy, Clone, Default)]
+    struct Wrapped<T: Default + Copy + std::fmt::Debug>(T);
+
+    impl MulAcc<Wrapped<i8>, Wrapped<i16>> for Wrapped<i32> {
+        fn mul_acc(&mut self, a: &Wrapped<i8>, b: &Wrapped<i16>) {
+            self.0 = self.0 + a.0 as i32 * b.0 as i32;
+        }
+    }
+
+    #[test]
+    fn mul_acc_mixed_param_sizes() {
+        let mut a = Wrapped::<i32>(0x40000007i32);
+        let b = Wrapped::<i8>(0x20i8);
+        let c = Wrapped::<i16>(0x3000i16);
+        a.mul_acc(&b, &c);
+        assert_eq!(a.0, 0x40060007i32);
     }
 }
--- a/src/sparse/prod.rs
+++ b/src/sparse/prod.rs
@@ -10,19 +10,43 @@ use num_traits::Num;
 /// Compute the dot product of two sparse vectors, using binary search to find matching indices.
 ///
 /// Runs in O(MlogN) time, where M and N are the number of non-zero entries in each vector.
-pub fn csvec_dot_by_binary_search<N, I>(
-    vec1: CsVecViewI<N, I>,
-    vec2: CsVecViewI<N, I>,
+pub fn csvec_dot_by_binary_search<N, I, A, B>(
+    vec1: CsVecViewI<A, I>,
+    vec2: CsVecViewI<B, I>,
 ) -> N
 where
     I: SpIndex,
-    N: crate::MulAcc + num_traits::Zero,
+    N: crate::MulAcc<A, B> + num_traits::Zero,
 {
-    let (mut idx1, mut val1, mut idx2, mut val2) = if vec1.nnz() < vec2.nnz() {
-        (vec1.indices(), vec1.data(), vec2.indices(), vec2.data())
+    // Check vec1.nnz<vec2.nnz
+    // Reverse the dot product vec1 and vec2, but preserve possibly non-commutative MulAcc
+    // through a lamba.
+    if vec1.nnz() > vec2.nnz() {
+        csvec_dot_by_binary_search_impl(vec2, vec1, |acc: &mut N, a, b| {
+            acc.mul_acc(b, a)
+        })
     } else {
-        (vec2.indices(), vec2.data(), vec1.indices(), vec1.data())
-    };
+        csvec_dot_by_binary_search_impl(vec1, vec2, |acc: &mut N, a, b| {
+            acc.mul_acc(a, b)
+        })
+    }
+}
+
+/// Inner routine of `csvec_dot_by_binary_search`, removes need for commutative `MulAcc`
+pub(crate) fn csvec_dot_by_binary_search_impl<N, I, A, B, F>(
+    vec1: CsVecViewI<A, I>,
+    vec2: CsVecViewI<B, I>,
+    mul_acc: F,
+) -> N
+where
+    F: Fn(&mut N, &A, &B),
+    I: SpIndex,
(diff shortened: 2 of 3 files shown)


mulimoen:
This looks great! Any clea rbenefit in using two type bounds in `MulAcc`? Is multiplying `i16` by `i32` and outputting `i64` something we should implement, or is it enough to have just `MulAcc<N>`?

It would be awsome to see this working with your use-case, could you make an impl in `src/mul_acc`and implement e.g. `MulAcc<i8, i8> for i16` to show the full benefit? Otherwise one must create a newtype if one wishes to use this due to orphan rules.

experiment9123:
thanks for reviewing the changes

> This looks great! Any clea rbenefit in using two type bounds in `MulAcc`? Is multiplying `i16` by `i32` and outputting `i64` something we should implement, or is it enough to have just `MulAcc<N>`?

My use case (explained below) definitely does need 2 types. they can be defaulted I think <SrcA=Self,SrcB=SrcA> if that seems friendlier.

bear in mind Dimension Checked maths aswell which some people like .. Rust now has const generics (yay!) making that possible. I've seen people write physics engines with that used all the way through

> It would be awsome to see this working with your use-case, could you make an impl in `src/mul_acc`and implement e.g. `MulAcc<i8, i8> for i16` to show the full benefit? Otherwise one must create a newtype if one wishes to use this due to orphan rules.

my use case is speculative , but let me elaborate:
[1] I wrote a "Graph<Node,Edge>" starting out on a spiking neuron framework experiment (trying to push rust on those peeps)
[2] i then realised the critical step of information flowing across edges was actually the  same pattern as Sparse Matrix X Sparse Vector. so I reworked this to use a mini sparse-matrix impl (COO format),then stored a  SparseMatrixCOO<Edge>, with Edge*Node producing the 'spike'. I was actually testing this update with Conways GoL for something more recognisable. I only did "SparseMatrix X Dense Vector" - and figured "ok i'll look for an existing matrix lib to do actual sparse vectors" - and sure enough I see SPRS appears to have this.
[3] I tried *Eigen* in C++ for the same idea and found they can indeed do it (you just need to supply some messy C++ style 'type traits')
[4] In graphics I have seen cases like 16bit matrix X 32bit coordinates, or 32bit matrices X 16bit compressed coords or 16x16bit but with different Fixed Point precisions. the permutaions that have appeared through the ages are endless.

So to get as far as using "sprs" in my use case , I will need to refactor all steps to get the matrix mul working. 
I estimate it's a few PRs (too much for one "sitting",reading code is hard) to get it done - I'd worry about doing it all in one go (the project can diverge, then you get merge hell, and it'll be harder to verify each change.)

the work I've done so far on it shows it's definitely possible, but it reuiqres a few reworks here and there (I saw another place where it looked like it did "to_owned()" to return an empty vector "if src was empty", like it was cloning the input, have to stop think  to get the type right there.. etc)

https://github.com/experiment9123/graphengine/blob/master/rustgraph/src/lib.rs
https://github.com/experiment9123/graphengine/blob/master/rustgraph/examples/gol.rs

src/mul_acc.rs
```diff
@@ -12,17 +12,21 @@
 /// to provide the most performant implementation. For instance, we could have
 /// a default implementation for numeric types that are `Clone`, but it would
 /// make possibly unnecessary copies.
-pub trait MulAcc {
+
+pub trait MulAcc<SrcA = Self, SrcB = SrcA> {
```
mulimoen, line 16: Remove newline, rename `SrcA` to `A` and likewise for B.

src/mul_acc.rs
```diff
@@ -12,17 +12,21 @@
...
     /// Multiply and accumulate in this variable, formally `*self += a * b`.
-    fn mul_acc(&mut self, a: &Self, b: &Self);
+    fn mul_acc(&mut self, a: &SrcA, b: &SrcB);
 }
 
-impl<N> MulAcc for N
+// impl for combinations of types which already support mul_add
+impl<AccN, SrcA, SrcB> MulAcc<SrcA, SrcB> for AccN
```
mulimoen, line 22: `AccN` -> `N`

src/sparse/prod.rs
```diff
@@ -10,19 +10,49 @@ use num_traits::Num;
...
+        return csvec_dot_by_binary_search_sub(
+            vec1,
+            vec2,
+            |acc: &mut N, a, b| acc.mul_acc(a, b),
+        );
+    }
+}
+
+// Inner routine of csvec_dot_by_binary_search, removes need for commutative MulAcc
+pub fn csvec_dot_by_binary_search_sub<N, I, A, B, F>(
```
mulimoen, line 42: `pub(crate)`

src/sparse/prod.rs
```diff
@@ -10,19 +10,49 @@ use num_traits::Num;
...
-    };
+        return csvec_dot_by_binary_search_sub(
+            vec1,
+            vec2,
+            |acc: &mut N, a, b| acc.mul_acc(a, b),
+        );
+    }
+}
+
+// Inner routine of csvec_dot_by_binary_search, removes need for commutative MulAcc
```
mulimoen, line 41: Triple `///` makes this into a doc comment

src/sparse/prod.rs
mulimoen: And name it `csvec_dot_by_binary_search_impl`

src/sparse/prod.rs
```diff
@@ -10,19 +10,49 @@ use num_traits::Num;
...
-    N: crate::MulAcc + num_traits::Zero,
+    N: crate::MulAcc<A, B> + num_traits::Zero,
 {
-    let (mut idx1, mut val1, mut idx2, mut val2) = if vec1.nnz() < vec2.nnz() {
-        (vec1.indices(), vec1.data(), vec2.indices(), vec2.data())
+    //Check vec1.nnz<vec2.nnz ; call self with args flipped if not
+    // can't work with the original method with seperate A,B types
+    // because the val1 / val2 would be swapped from &[A]<->&[B]
+    // lambda passed into the subroutine dodges the need for MulAcc<B,A>
+    // we might want to fit non-commutative plugin MulAcc functions, eg 'vectors of small dense matrices' etc.
```
mulimoen, line 25: Maybe better phrased as:
Optimisation: Reverse the dot product `vec1` and `vec2`, but preserve possibly non-commutative `MulAcc` through a lamba.

mulimoen: Would be interesting to see if the performance is improved by using `NnzOrZip` in this case, so we don't have to binary search one array all the time.

mulimoen (commented):
Some smaller style fixes, otherwise looks good

src/sparse/prod.rs
experiment9123: right all these style fixes are fine by me.  foo_impl is a better name for 'foo_sub(routine)'  _sub in a maths lib is indeed confusing (i was also tempted to make it a local lambda but its too much nesting)

Are you integrating these tweaks directly or do you want me to replicate them prior to the pull request being approved?

mulimoen: I can fix them this afternoon, but you may also fix them if you prefer

mulimoen:
If you rebase on master you'll get the CI fixes. If you could add a test where you use the `MulAcc` trait on a newtype, with A/=B/=N would be great for this PR

experiment9123:
> If you rebase on master you'll get the CI fixes. If you could add a test where you use the `MulAcc` trait on a newtype, with A/=B/=N would be great for this PR

ok i've added a test like this - mul_acc_mixed_param_sizes  ,     impl MulAcc<Wrapped<i32>, Wrapped<i16>> for Wrapped<i64> {... (int values for precise equality test)

src/mul_acc.rs
```diff
@@ -38,4 +42,22 @@ mod tests {
...
+        fn mul_acc(&mut self, a: &Wrapped<i32>, b: &Wrapped<i16>) {
+            self.0 = self.0 + a.0 as i64 * b.0 as i64;
+        }
+    }
+
+    #[test]
+    fn mul_acc_mixed_param_sizes() {
+        let mut a = Wrapped::<i64>(1i64);
+        let b = Wrapped::<i32>(2i32);
+        let c = Wrapped::<i16>(3i16);
```
mulimoen, line 59: Use `i16/i32::max_value()` to show why we need the trait

experiment9123: could you elaborate (use 'max_value' as the example vals, ie can't accumulate into their values, or as assertions perhaps?)
another idea might be 0x2000i16, 0x30000000i32  , mul-accumulating into the i64 (recognizable values, with products that dont fit). i might also change the test to i32+=i8*i16 for that (easier to see)

mulimoen: It's enough to have something which would overflow, with max_value this would be obvious, but your choices here are good. Just make sure the sum would overflow all but the output type

experiment9123: ok it now does '0x1i32 + 0x20i8 * 0x3000i16 = 0x70001i32 in the test

mulimoen:
I took the liberty of rebasing and fixing some minor nits. Thank you @experiment9123 for your work on this! I'm excited to see this implemented on `Mul` for matrices.

experiment9123:
thats great.. i'll grab the latest version and try and make my way back through the callgraph till we can use this in sparse matrix mul X sparse vec.. i suspect it'll be several PR's yet, but we might get it for a user-facing "sparse dot product" first..

PR #288, opened 2021-04-26
added MulAcc<A,B> support for matrix csc & csr X dense vec formats

experiment9123:
2 functions that were easy to do. 1 minor change in each - flip the mul_acc operands to be matrix elem, vec elem (the order you see it in the expression Matrix X Vector) - and the tests ended up needing a couple of type annotations. besides that the callgraph is unaffected, everywhere else those fn's are used.. nothing changed.

### DIFF
--- a/src/sparse/prod.rs
+++ b/src/sparse/prod.rs
@@ -70,15 +70,15 @@ where
 
 /// Multiply a sparse CSC matrix with a dense vector and accumulate the result
 /// into another dense vector
-pub fn mul_acc_mat_vec_csc<N, I, Iptr, V, VRes>(
-    mat: CsMatViewI<N, I, Iptr>,
+pub fn mul_acc_mat_vec_csc<N, I, Iptr, V, VRes, A, B>(
+    mat: CsMatViewI<A, I, Iptr>,
     in_vec: V,
     mut res_vec: VRes,
 ) where
-    N: crate::MulAcc,
+    N: crate::MulAcc<A, B>,
     I: SpIndex,
     Iptr: SpIndex,
-    V: DenseVector<Scalar = N>,
+    V: DenseVector<Scalar = B>,
     VRes: DenseVectorMut<Scalar = N>,
 {
     let mat = mat.view();
@@ -90,25 +90,25 @@ pub fn mul_acc_mat_vec_csc<N, I, Iptr, V, VRes>(
     }
 
     for (col_ind, vec) in mat.outer_iterator().enumerate() {
-        let multiplier = in_vec.index(col_ind);
-        for (row_ind, value) in vec.iter() {
+        let vec_elem = in_vec.index(col_ind);
+        for (row_ind, mtx_elem) in vec.iter() {
             // TODO: unsafe access to value? needs bench
-            res_vec.index_mut(row_ind).mul_acc(multiplier, value);
+            res_vec.index_mut(row_ind).mul_acc(mtx_elem, vec_elem);
         }
     }
 }
 
 /// Multiply a sparse CSR matrix with a dense vector and accumulate the result
 /// into another dense vector
-pub fn mul_acc_mat_vec_csr<N, I, Iptr, V, VRes>(
-    mat: CsMatViewI<N, I, Iptr>,
+pub fn mul_acc_mat_vec_csr<N, A, B, I, Iptr, V, VRes>(
+    mat: CsMatViewI<A, I, Iptr>,
     in_vec: V,
     mut res_vec: VRes,
 ) where
-    N: crate::MulAcc,
+    N: crate::MulAcc<A, B>,
     I: SpIndex,
     Iptr: SpIndex,
-    V: DenseVector<Scalar = N>,
+    V: DenseVector<Scalar = B>,
     VRes: DenseVectorMut<Scalar = N>,
 {
     if mat.cols() != in_vec.dim() || mat.rows() != res_vec.dim() {
@@ -120,9 +120,9 @@ pub fn mul_acc_mat_vec_csr<N, I, Iptr, V, VRes>(
 
     for (row_ind, vec) in mat.outer_iterator().enumerate() {
         let tv = res_vec.index_mut(row_ind);
-        for (col_ind, value) in vec.iter() {
+        for (col_ind, mtx_elem) in vec.iter() {
             // TODO: unsafe access to value? needs bench
-            tv.mul_acc(in_vec.index(col_ind), value);
+            tv.mul_acc(mtx_elem, in_vec.index(col_ind));
         }
     }
 }
@@ -396,7 +396,7 @@ mod test {
         assert!(res_vec
             .iter()
             .zip(expected_output.iter())
-            .all(|(x, y)| (*x - *y).abs() < epsilon));
+            .all(|(x, y): (&f64, &f64)| (*x - *y).abs() < epsilon));
     }
 
     #[test]
@@ -446,7 +446,7 @@ mod test {
         assert!(res_vec
             .iter()
             .zip(expected_output.iter())
-            .all(|(x, y)| (*x - *y).abs() < epsilon));
+            .all(|(x, y): (&f64, &f64)| (*x - *y).abs() < epsilon));
     }
 
     #[test]


mulimoen:
I would like to see more of the code converted, just to see if type inference still works. Could you implement equally for all the `smmp` code too? This would make porting `Mul<CsMat>` trivial, so you could add this one as well.

experiment9123:
now I tried converting some more, and there is one thing where the inference starts asking for ascriptions here and there.
in some of the tests, the result type is not inferable from the inside; hence I had to modify them eg "assert_eq!({let x:f32=blah;x},expected)

the 'examples' code is unchanged.

Nonetheless if these needed ascriptions seem like a PITA,
one idea would be to rename these fully generic "dot" "dot_infered" or something, and have a second "dot" with the original type assumptions.

Take a look at the measures I had to take in the latest version, and see what you think.

experiment9123:
> smmp code too? This would make porting Mul<CsMat> trivial, so you could add this one as well.

Now locally, I tried converting `Mul<CsMat>` with  the usual idea of N=output types, A,B=inputs (type Output=CsMat<N> yada yada), and it seems to say "type param N is unconstrained" . so ..hmmm.. i'm unsure if the fully generic case can work (although I had assumed it should??) . 

trying to think why this would be any different to the dot-product cases. Possibly because they had the return type params *in the fn def* , wheras here it's an associated type.

Perhaps we could just keep the operators using a single type, and have a named function to do this job?

experiment9123:
I did another little experiment (not pushed) where I made a helper trait A:MulAddWith<B>{ type Accumulator } .. with impl such that A:MulAddWith<B, Accumulator=N> - allows binary operators to express equiv of c++ typedef decltype(A()*B()+A()*B()) N , "N=type_of(A*B+A*B)" 

one fly in the ointment with that is the default impls might have conflicted with autoimpling MulAdd ->MulAcc (i think) i.e. once you have both A:MulAddWith<B>, N:MulAcc<A,B> yields the conflict with MulAdd

if this  is the case, one workaround might be to manually impl the MulAccs using MulAdds for specific types in a macro (the primitive ints and floats)

i'm trying to think if I could build the MulAdd requirement into the MulAddWith trait .. but in the meantime i've converted the other easy functions

experiment9123:
other ideas

// keep '*' fixed types for the time-being
```

let output:CsVec<MyAcc> = my_sparse_mat.lazy_mul_csvec(&my_sparse_vec).into() 
   fn lazy_mul_cs_vec(&self:CsMat<A,..>  , rhs:&CsVec<B,...)->LazyProduct< CsMat<A>,CsVec<B>>
   impl<N,A,B> Into<CsVec<N>> for LazyProduct<...>

// inner implemenatino of Mul<CsMat..> for CsMat<..> and LazyProduct<>.into() shared in another "named function"

   fn csmat_mul_csvec_into<A,B,N>(dst_vec:&mut CsVec<>,  src_mat:&CsMat<A,..> , src_vec:&CsVec<..>){
     // common code extracted from Mul for CsMat...
   }




```

experiment9123:
right in the latest commit, i've extracted the Operator * guts into `pub fn csmat_mul_csmat`, which seems to work fine with fully generic inputs/outputs, `N:MulAcc<A,B>` . I'm uneasy about the idea of using `Mul, Add Output`'s to compute it because these aren't related to a generic impl for MulAcc.  the free function means the functionality is *available*, but less sadly discoverable. 

perhaps we can get the functionality and internals all working fine before we attack ideas like Lazy Expression templates ( I would also suggest that a "member function" could be almost as discoverable as the operator.)

mulimoen:
The extra type annotations are a bit worrying, and would make the code harder to use for first time users. Adding some free-standing functions which uses `MulAcc` internally would be a good first step, which is what we have in the `smmp` module. Maybe adding these accumulating methods on `CsMat`/`CsVec` is good enough for this specialized use, and would leave `Mul` for the scalar case?

mulimoen:
For this case we must ensure functions requireing `MulAcc` takes care of the non-commutate operation. Are there no way to fix this as was done in #288?

experiment9123:
> The extra type annotations are a bit worrying
so far i've only seen it affect (i) the tests and (ii) benchmarks - which dont store or return anything .. i'm hoping it'll "just work"
and these tests use the "internal" named functions. Currently all the operators still do N:MulAcc<N,N>, so their interface is unchanged.

> Maybe adding these accumulating methods on CsMat/CsVec is good enough for this specialized use

I would certainly be happy with that. what I'm also seeing is that even "returning" named functions infer their types better. I think it's that the associated types in the operators is just slightly different to having a type-param list on the function, even though there's no difference in principle (eg the new func csmat_mul_csmat infers N,A,B, just fine). it *should* just be like "a function that happens to be called `*`".. but isn't exactly.

>For this case we must ensure functions requireing MulAcc takes care of the non-commutate operation. Are there no way to fix this as was done in #288?

what I've seen of the code so far .. it should _definitely_ be possible but it's intense to figure out the detail, i'll need to take a breather: the issue is the matrix ops flip the args and call multiples with transposed versions of themselves, and I think there's several places I'd need to add and propogate some kind of argflip lambda.

after a pause and learning abit more about this source base , new approaches might click . 

If the layouts were compile-time known (like explicit struct SparseMatrixCsc<T>, SparseMatrixCsr<T> etc)  the argflip problem goes away. but I'm guessing it does this because some of those permuations ("row vs col major" x "csc/csr") are just equivalent to eachother (what I have in my head is seperate types, with simpler Mul overload implementations, then an uber-type which wraps them in an enum)


in the use case I personally have in mind, it's going to be a known format (csr or csc) X sparse vector - there's probably enough ported already to handle this now (I can call the format-specific functions). 


A couple of motivations here.. the "experiment I had in mind" is pretty much doable now, but I also care about the "reputation of the language & ecosystem". compare the capabilities of Eigen/C++ with "the best sparse matrix lib I can find in rust" .. it would be nice to have something as versatile as possible.

experiment9123:
been paused on this for a couple of days.. in 2 minds about it
if you merge this - you enable my use case (via fn csmat_mul_csmat) - but it does complicate the internals for the next person who tries to add a feature. The trait bounds get quite fiddly.

i'm increasingly tempted to just write the path+format i was after independently.. which is a shame as it doesn't feed back into the community as much as improving sprs. (it would take about 500-1000 lines max).

Haven't given up 100% though. This library could definitely do it, and it would be great to have something this versatile and compreheensive (to compete with C++ Eigen..)

src/sparse/csmat.rs
```diff
@@ -1795,24 +1795,63 @@ where
...
-                let rhs_csc = rhs.to_other_storage();
-                smmp::mul_csr_csr(
-                    rhs_csc.transpose_view(),
-                    self.transpose_view(),
-                )
+        csmat_mul_csmat(self, rhs)
+    }
+}
+
+/// multiply `CsMatBase<A>` with `CsMatBase<B>` with any types satisfying `N:MulAcc<A,B>`
```
mulimoen, line 1802: Suggestiong for comment here:
```rust
/// Multiply two sparse matrices.

/// This function is generic over `MulAcc`, and supports accumulating into a different output type.
/// This is not the default for `Mul`, as type inference fails for intermediaries
```

src/sparse/prod.rs
```diff
@@ -120,51 +120,60 @@ pub fn mul_acc_mat_vec_csr<N, I, Iptr, V, VRes>(
...
+    N: crate::MulAcc<A, B> + num_traits::Zero + PartialEq + Clone,
+    A: Copy + num_traits::Zero,
+    B: Copy + num_traits::Zero,
     I: SpIndex,
     Iptr: SpIndex,
 {
     if rhs.dim == 0 {
-        return rhs.to_owned();
+        // create an empty sparse vector of correct dimension
+        return CsVecI::<N, I>::empty(0);
```
mulimoen, line 176: `return Self::empty(0)`

src/sparse/smmp.rs
```diff
@@ -190,12 +192,14 @@ pub fn numeric<
...
-    rhs: CsMatViewI<N, I, Iptr>,
+pub fn mul_csr_csr<N, A, B, I, Iptr>(
+    lhs: CsMatViewI<A, I, Iptr>,
+    rhs: CsMatViewI<B, I, Iptr>,
 ) -> CsMatI<N, I, Iptr>
 where
-    N: crate::MulAcc + num_traits::Zero + Clone + Send + Sync,
+    N: crate::MulAcc<A, B> + num_traits::Zero + Clone + Send + Sync,
+    A: num_traits::Zero + Send + Clone + Sync,
+    B: num_traits::Zero + Send + Clone + Sync,
```
mulimoen, line 202: These two only needs to be `Send + Sync`, we never do anything but read them.

src/sparse/smmp.rs
```diff
@@ -246,14 +250,16 @@ where
...
+    lhs: CsMatViewI<A, I, Iptr>,
+    rhs: CsMatViewI<B, I, Iptr>,
     seens: &mut [Box<[bool]>],
     tmps: &mut [Box<[N]>],
 ) -> CsMatI<N, I, Iptr>
 where
-    N: crate::MulAcc + num_traits::Zero + Clone + Send + Sync,
+    N: crate::MulAcc<A, B> + num_traits::Zero + Clone + Send + Sync,
+    A: num_traits::Zero + Clone + Send + Sync,
+    B: num_traits::Zero + Clone + Send + Sync,
```
mulimoen, line 262: Ditto, `Send + Sync`

src/sparse/vec.rs
```diff
@@ -821,25 +821,25 @@ where
     /// assert_eq!(4., v1.dot(&v1));
     /// assert_eq!(16., v2.dot(&v2));
     /// ```
-    pub fn dot<'b, T: IntoSparseVecIter<'b, N>>(&'b self, rhs: T) -> N
+    pub fn dot<'b, T: IntoSparseVecIter<'b, X>, X, Y>(&'b self, rhs: T) -> Y
```
mulimoen, line 824: This break type inference, as we see in our tests. Could we introduce `dot_acc` as an accumulating dot product and leave `dot` as `MulAcc<N, N>`?

src/sparse/vec.rs
```diff
@@ -821,25 +821,25 @@ where
...
     where
-        N: 'b + crate::MulAcc + num_traits::Zero,
-        I: 'b,
-        <T as IntoSparseVecIter<'b, N>>::IterType:
-            Iterator<Item = (usize, &'b N)>,
+        Y: 'b + crate::MulAcc<N, X> + num_traits::Zero,
+        X: 'b + Copy + num_traits::Zero,
```
mulimoen, line 827: `X` does not need `Copy` nor `Zero`

src/sparse/vec.rs
```diff
@@ -1086,6 +1086,7 @@ impl<'a, 'b, N, I, Iptr, IpS1, IS1, DS1, IS2, DS2>
     for &'a CsMatBase<N, I, IpS1, IS1, DS1, Iptr>
 where
     N: Clone
+        + Copy
```
mulimoen, line 1089: Why do you add `Copy` here?

mulimoen (changes requested):
Thanks for your patience, I've been rather busy this week. I like most of these changes, it does work exactly as intended, except type inference failing occasionally. If we rename/replace `dot` we can retain backwards compatability, and still have access to `MulAcc` when necessary.

mulimoen:
> If the layouts were compile-time known

I think this would be useful for reasoning about the cost of matrix operations, but it would possibly explode the number of `Mul` impls. How would this interact with `CsMat`, `CsMatI`, `CsMatView`, `CsMatViewMut` etc.?

experiment9123:
> > If the layouts were compile-time known
> 
> I think this would be useful for reasoning about the cost of matrix operations, but it would possibly explode the number of `Mul` impls. How would this interact with `CsMat`, `CsMatI`, `CsMatView`, `CsMatViewMut` etc.?

How it would interact with all those types - I dont even know my way around them fully enough to definitively answer that. You're right it's a concern.

I would not suggest supporting all of them - only a few permtations that have particular reason to exist (starting with my own particular use case) .
I need to draw a roadmap ..

One possibility would be:
[Step 1] - Operators only work with uniform types.   Named functions for the multi-type generic matrix mul. Expose some internals for known type combinations,

[Step 2] expose a few known, easy combinations with operators for the multi-type cases.     eg..   DenseVec X SparseMatrixCOO -> DenseVec has an easy impl.   SparseVec X MatrixCSC -> DenseVec has an easy "Scatter" style multiply impl.     DenseVec X MatixCSR has an easy "Gather" style multiply impl.     ( for my use case, i''m guessing that SparseVec X MatrixCSC->Dense -> re-sparsify the result is worth trying)

[Step 3] if we fix "the param flip problem" the fully generic, all-types approach can be fully generalised.

src/sparse/vec.rs
experiment9123: right someimtes I can't even quite follow exactly when it needs Copy or not. I was just fixing compile errors. I figure there might be scope for a T:MatrixElem kind of shared bound to put in these places (do we need Clone / Zero / Copy or what by default).   It did seem like with MulAcc::mul(&a,&b)  "copy" shouldn't be needed. .. but sometimes maths libs make an assumption that the component types are trivially copyable (I am writing heavily generic 'graphics' type vecmaths myself elsewhere, i've bounced back and forth on this myself)

let me go over some of the other reqested changes.. were you able to integrate them yourself (then I can grab them) or shall i replicate them ? I'm happy with all those updates myself.

> This break type inference, as we see in our tests. Could we introduce dot_acc as an accumulating dot product and leave dot as MulAcc<N, N>?

I'll have to think more about this possiblitiy .. do you mean a seperate   `my_accum_vec.dot_acc(&vec1,&vec2) /* my_accum_vec += dot(vec1,vec2)`  .. which would obviously be easier for inference?  certainly seems merit in something like that, along with maybe even "accumulating matrix multiplies" aswell

experiment9123:
(just been reading through the requested changes, i'm ok with the suggestions, i'll have to figure out exactly what was happening with dot and the suggestion there .. are you suggesitng reverting some of the generics that break type inference for existing users, then adding *new* functions like "dst.dot_acc(&src1,&src2)" for full inference  ? i'm fine  with anything like that

mulimoen:
Build failure is due to `gitlab` upgrade, will rerun and merge later on. Thanks @experiment9123 for this PR!

experiment9123 (commented):
fine by me

experiment9123:
Hi just letting you know I havent forgotten about this, i have paused whilst working on other things (still very much in rust).
let me figure out how to grab your suggested changes. there is a risk that if i dont get this done this "round" of changes will diverge.

I agree it really wants the "lambda parameter flip" for the non-commutative behaviour for 100% versatility

PR #290, opened 2021-07-05
Add asymmetric operands to pointwise binary ops

tchernobog:
Alters the existing pointwise addition, subtraction, and multiplication binary operators so that they allow for different left-hand and right-hand side operands.

Note that in some cases this can lead to minor backwards incompatibilities due to the need to specify the type of one operand as it cannot be inferred automatically (e.g. when using an "eye" matrix). Please consider bumping the minor version number if merging.

### DIFF
--- a/src/sparse/binop.rs
+++ b/src/sparse/binop.rs
@@ -275,22 +275,35 @@ where
 /// The matrices must have the same ordering, a `CSR` matrix must be
 /// added with a matrix with `C`-like ordering, a `CSC` matrix
 /// must be added with a matrix with `F`-like ordering.
-pub fn add_dense_mat_same_ordering<Lhs, Rhs, Res, I, Iptr, Mat, D>(
+pub fn add_dense_mat_same_ordering<
+    Lhs,
+    Rhs,
+    Res,
+    Alpha,
+    Beta,
+    ByProd1,
+    ByProd2,
+    I,
+    Iptr,
+    Mat,
+    D,
+>(
     lhs: &Mat,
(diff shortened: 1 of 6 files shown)


mulimoen (commented):
This looks good, and fits well in with the other generalisations made in this crate since the last release. It is not ideal to break type inference for existing code, although this is a minor cost for this additional functionality. We have already accumulated some breaking changes, so this is OK.

Could you add a couple of tests illuminating the added functionality? A couple of newtypes wrapping e.g. `i16 + i32 = i64` would be nice.

tchernobog:
Done. Please note that the current incarnation only implements vector addition and subtraction. The other operations, and extending to also cover matrices, can be added if there is interest. I didn't need it right away for my own project, so I held off for now.

src/sparse/binop.rs
```diff
@@ -351,11 +420,11 @@ pub fn csmat_binop_dense_raw<'a, N, I, Iptr, F>(
             .iter_mut()
             .zip(rrow.iter().enumerate().nnz_or_zip(lrow.iter()))
         {
-            let (oval, lr_elems) = items;
-            let binop_val = match lr_elems {
-                Left((_, val)) => binop(val, &N::zero()),
-                Right((_, val)) => binop(&N::zero(), val),
-                Both((_, lval, rval)) => binop(lval, rval),
```
mulimoen, line 358: Well spotted on this weird iteration

mulimoen (approved):

mulimoen:
Thanks for your great work on this @tchernobog!

PR #292, opened 2021-07-31
Fix regression in type resolution

mulimoen:


mulimoen:
@bytesnake If you could confirm this fixes the compilation error I'll merge and release a new version on `crates.io`

PR #304, opened 2022-06-21
minor code fix to upgrade pyo3 to 0.16.5

chetmurthy:
I'm not sure if this is done correctly, but it seems to work (passes cargo test, cargo bench).

It seems the way a method gets called has changed.


mulimoen:
Thanks @chetmurthy for this bump!

PR #309, opened 2022-06-26
minor refactor, to support reading/writing to/from gzip files

chetmurthy:
Made a tiny refactor, in order to support I/O to gzipped matrixmarket files.

Also changed trait bound to relax ?Sized on BufRead/Write, for
same end-goal.


mulimoen:
Thanks!

PR #316, opened 2022-09-06
Derive Clone for VectorIterator

RobinStoll:
This adds deriving of `Clone` for `VectorIterator`. This is relevant e.g. for [`multi_cartesian_product`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.multi_cartesian_product) from `itertools`.


mulimoen:
Thanks @RobinStoll!

Issue #318, opened 2022-09-13
Add support for pattern type


shenjiangqiu:
according to https://math.nist.gov/MatrixMarket/formats.html#MMformat, there are three data types supported by the MTX file: num, complex, and pattern. Implementing a pattern type is quite easy under the current code.
I did some job for that, is that seem good?

PR #319, opened 2022-09-13
update pattern type

shenjiangqiu:
the [pattern]( https://math.nist.gov/MatrixMarket/formats.html#MMformat) type is simply a zero-sized structure. update the io module to support:
1. read an MTX file with pattern type
2. read a normal non-pattern MTX file as a pattern-type MTX, ignore the data filed. #318

### DIFF
--- a/sprs/src/io.rs
+++ b/sprs/src/io.rs
@@ -826,6 +826,7 @@ mod test {
     }
 
     #[test]
+    #[cfg_attr(miri, ignore)]
     fn read_pattern_matrix() {
         let path = "data/matrix_market/pattern.mm";
         let mat = read_matrix_market::<Pattern, usize, _>(path).unwrap();
@@ -856,6 +857,7 @@ mod test {
     }
 
     #[test]
+    #[cfg_attr(miri, ignore)]
     fn read_non_pattern_mtx_to_pattern() {
         let path = "data/matrix_market/simple.mm";
         let mat = read_matrix_market::<Pattern, usize, _>(path).unwrap();
--- a/sprs/src/io.rs
+++ b/sprs/src/io.rs
@@ -857,7 +857,6 @@ mod test {
         let mat = read_matrix_market::<Pattern, usize, _>(path).unwrap();
         println!("trimat: {:?}", mat);
         let csc = mat.to_csc();
-        assert_eq!(csc.get(0, 0), Some(&Pattern {}));
         assert_eq!(csc.get(1, 1), Some(&Pattern {}));
         assert_eq!(csc.get(2, 2), Some(&Pattern {}));
         assert_eq!(csc.get(3, 3), Some(&Pattern {}));
@@ -879,5 +878,20 @@ mod test {
         write_matrix_market(&save_path, &csc).unwrap();
         let mat2 = read_matrix_market::<Pattern, usize, _>(&save_path).unwrap();
         assert_eq!(csc, mat2.to_csc());
+    }
+
+    #[test]
+    #[cfg_attr(miri, ignore)]
+    fn read_csc_to_csr_trans() {
+        let path = "data/matrix_market/simple.mm";
+        let mat = read_matrix_market::<Pattern, usize, _>(path).unwrap();
+        println!("trimat: {:?}", mat);
+        let csc: CsMat<Pattern> = mat.to_csc();
+        let csr = csc.to_csr();
+        let csc_copy = csr.to_csc();
+        let csr_copy = csc_copy.to_csr();
+
+        assert_eq!(csc, csc_copy);
+        assert_eq!(csr, csr_copy);
     }
 }
--- a/sprs/src/io.rs
+++ b/sprs/src/io.rs
@@ -5,9 +5,8 @@ use std::fmt;
 use std::fs::File;
 use std::io;
 use std::io::{Seek, SeekFrom, Write};
+use std::ops::Neg;
 use std::path::Path;
-
-use num_traits::cast::NumCast;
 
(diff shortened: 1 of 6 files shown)


sprs/src/io.rs
```diff
@@ -251,8 +260,19 @@ where
         if sym_mode == SymmetryMode::SkewSymmetric && row == col {
             return Err(BadMatrixMarketFile);
         }
-        if entry.next().is_some() {
-            return Err(BadMatrixMarketFile);
+        match droping_data {
```
mulimoen, line 263: Replace `match` by `if`

sprs/src/io.rs
```diff
@@ -802,4 +824,66 @@ mod test {
...
+    #[cfg_attr(miri, ignore)]
+    fn read_pattern_matrix() {
+        let path = "data/matrix_market/pattern.mm";
+        let mat = read_matrix_market::<Pattern, usize, _>(path).unwrap();
+        println!("trimat: {:?}", mat);
+        let csc = mat.to_csc();
+        assert_eq!(csc.get(0, 0), Some(&Pattern {}));
+        assert_eq!(csc.get(1, 1), Some(&Pattern {}));
+        assert_eq!(csc.get(2, 2), Some(&Pattern {}));
+        assert_eq!(csc.get(3, 3), Some(&Pattern {}));
```
mulimoen, line 838: Replace `Pattern {}` by `Pattern` (drop braces)

sprs/src/num_kinds.rs
```diff
@@ -2,13 +2,106 @@
...
+    fn add(self, _other: Pattern) -> Pattern {
+        Pattern {}
+    }
+}
+impl Zero for Pattern {
+    fn zero() -> Self {
+        Pattern {}
+    }
+    fn is_zero(&self) -> bool {
+        true
```
mulimoen, line 25: This should be `false` as absence marks a zero

sprs/src/num_kinds.rs
```diff
@@ -2,13 +2,106 @@
...
+
+impl Add for Pattern {
+    type Output = Pattern;
+    fn add(self, _other: Pattern) -> Pattern {
+        Pattern {}
+    }
+}
+impl Zero for Pattern {
+    fn zero() -> Self {
+        Pattern {}
```
mulimoen, line 22: `Self`

sprs/src/num_kinds.rs
```diff
@@ -2,13 +2,106 @@
...
+    }
+    fn is_zero(&self) -> bool {
+        true
+    }
+
+    fn set_zero(&mut self) {
+        *self = Zero::zero();
```
mulimoen, line 29: Maybe panic instead?

sprs/src/num_kinds.rs
```diff
@@ -2,13 +2,106 @@
...
+    }
+}
+impl Rem for Pattern {
```
mulimoen, line 32: I am not sure why we need these traits, they make little sense for a pattern type

sprs/src/io.rs
```diff
@@ -160,10 +163,16 @@ where
         DataType::Integer => NumKind::Integer,
         DataType::Real => NumKind::Float,
         DataType::Complex => NumKind::Complex,
+        DataType::Pattern => NumKind::Pattern,
     };
-    if N::num_kind() != data_num_kind {
+    // any type can be convert to pattern
```
mulimoen, line 168: Do you intend for float matrices to be interpreted as a pattern matrix? I would prefer if all such conversions were done explicitly by the user by e.g. mapping after reading.

mulimoen (commented):
Great to see pattern types being supported! Only some minor comments, this looks very promising.

sprs/src/num_kinds.rs
shenjiangqiu: Currently the to_csr and to_csc method needs <N> to be Num. Those traits are required to implement Num.
But in to_csr and to_csc, only Add and Zero are used. I suggest removing the Sum trait requirement for N. We can use N: Add<Output=N> instead. please check the newest Commit for detailed changes.

sprs/src/io.rs
shenjiangqiu: My consideration is that sometimes we have a lot of matrices in a folder, and we only care about the indices. Those matrices all have different types of data. It's hard for a user to read them all up and then convert them to the pattern.

sprs/src/num_kinds.rs
shenjiangqiu: The removal of the Num Trait .and then adding a new Add trait doesn't import break change. Cause all types implement Num must already implement Add.

sprs/src/num_kinds.rs
shenjiangqiu: I have deleted this trait.

sprs/src/num_kinds.rs
shenjiangqiu: I have deleted this trait.

sprs/src/num_kinds.rs
shenjiangqiu: I have deleted this trait.

sprs/src/io.rs
shenjiangqiu: because the user can't know the real type before an error is reported.

shenjiangqiu: I have an idea. We can add a new function like `read_...._ignore_data(...)`, the old read function behavior like before(will return an error if the matrix is not originally Pattern), and the new `read_...._ignore_data(...)` always return CSMat<Pattern>, does this sound better?

PR #320, opened 2022-12-28
relax trait requirements for binary operations.

taketo1024:
Following #319, I have relaxed the trait requirements for `Add`, `Sub` etc for `CsVec`, so that in particular we can add and subtract vectors with scalar types that does not necessarily implement the `num_traits::Num` trait.


taketo1024:
Sorry for the late reply.

> Would you mind rebasing on master to include the clippy lint fixes?
Sure!

taketo1024:
I have rebased `relax_trait_restr` on `master` and checked that test runs ok in `sprs`.

mulimoen:
Thanks for this PR @taketo1024!

Issue #323, opened 2023-01-18
Modifying the default implementation for `MulAcc`. 


taketo1024:
I'd like to use `sprs` with [`num-bigint`](https://github.com/rust-num/num-bigint), [`num-rational`](https://github.com/rust-num/num-rational) etc. However, `sprs`'s matrix multiplication requires that the scalar type implements `MulAcc`, which `BigInt` and `Rational` types do not suffice. 

We can make these types automatically implement `MulAcc`, if we change the code in [`mul_acc.rs`](https://github.com/sparsemat/sprs/blob/master/sprs/src/mul_acc.rs) from the current

```rust
impl<N, A, B> MulAcc<A, B> for N
where
    N: Copy,
    B: Copy,
    A: num_traits::MulAdd<B, N, Output = N> + Copy,
{
    fn mul_acc(&mut self, a: &A, b: &B) {
        *self = a.mul_add(*b, *self);
    }
}
```

to

```rust
impl<N, A, B> MulAcc<A, B> for N
where
    for<'x> &'x A: Mul<&'x B, Output = N>,
    N: AddAssign<N> 
{
    fn mul_acc(&mut self, a: &A, b: &B) {
        self.add_assign(a * b);
    }
}
```

Is this a bad idea?  (regarding performance?)
Thanks.

taketo1024:
I have pushed the modified version on my folked repo. https://github.com/taketo1024/sprs/tree/mul_acc_default. 
I will send a PR if this modification seems good.

mulimoen:
That seems like a good idea. We can judge the performance impact once the PR is up

taketo1024:
Created PR. Closing this issue. Thanks!

PR #324, opened 2023-01-19
Modifying the default implementation for MulAcc.

taketo1024:
Modified the default implementation of `MulAcc`, so that the default implementation applies to scalar types that does not necessary implement `Copy` and `num_traits::MulAdd`. In particular, this modification enables [`BigInt`](https://github.com/rust-num/num-bigint) and [`Rational`](https://github.com/rust-num/num-rational) types applicable. 

Follows discussion from #323.


mulimoen:
Thanks @taketo1024 !

PR #327, opened 2023-04-17
Add bindings to UMFPACK

jlogan03:
* Add build, raw bindings, and mid-level wrapper for i32-indexed and i64-indexed, real-valued versions of UMFPACK (the DI and DL variations)
    * Same footprint of bindings as scipy and Julia, I believe
* Just covers the most-traveled parts (opaque LU factorization, solution of Ax=b problem type, and recovery of LU factorization components)
* Unlike the other suitesparse integrations, this one only supports dynamic linking
    * The build for UMFPACK depends on several of the other packages plus BLAS, and while it's definitely possible to do it, it looks extremely time-consuming and I think it's best to leave that for another time


jlogan03:
This is ready for another CI pass whenever - format, doc, clippy, and test steps are all running locally now

jlogan03:
The functions that clippy didn't like on the last pass weren't used anymore and have been removed entirely.

jlogan03:
Looks like the static build is also going to need some work in the future and I'm just going to remove it for now. It looks extremely involved & nominal feature flag configuration doesn't appear to allow any of the static version to actually get built anyway

jlogan03:
Alrighty, ready for another pass of CI & passing lint/test locally

jlogan03:
Resolved the latest round of errors & ready for another go

suitesparse_bindings/sprs_suitesparse_umfpack/src/lib.rs
```diff
@@ -0,0 +1,380 @@
...
+     $free_numeric: ident,
+     $free_symbolic: ident,
+     ) => {
+
+        /// Wrapper of raw handle to guarantee proper drop procedure
+        struct $Symbolic(*mut c_void);
+
+        impl Drop for $Symbolic {
+            fn drop(&mut self) {
+                unsafe {$free_symbolic(&mut self.0 as *mut *mut c_void);}
```
mulimoen, line 33: `unsafe {$free_symbolic(std::ptr::addr_of_mut!(self.0).cast());}`

suitesparse_bindings/sprs_suitesparse_umfpack/src/lib.rs
```diff
@@ -0,0 +1,380 @@
...
+            }
+        }
+
+        /// Wrapper of raw handle to guarantee proper drop procedure
+        struct $Numeric(*mut c_void);
+
+        impl Drop for $Numeric {
+            fn drop(&mut self) {
+                unsafe {$free_numeric(&mut self.0 as *mut *mut c_void);}
```
mulimoen, line 42: Same as for `free_symbolic`

suitesparse_bindings/sprs_suitesparse_umfpack/src/lib.rs
```diff
@@ -0,0 +1,380 @@
...
+            }
+        }
+
+        /// Partial interface to `SuiteSparse`'s UMFPACK solver package.
+        /// Provides LU factorization of A and solution of Ax=b using stored factorization.
+        pub struct $Context {
+            /// `A` matrix of system Ax=b
+            _a: CsMatI<f64, $int>,
```
mulimoen, line 50: Drop underscore

suitesparse_bindings/sprs_suitesparse_umfpack/src/lib.rs
```diff
@@ -0,0 +1,380 @@
...
+                        ncol_ as $int,
+                        ap,
+                        ai,
+                        ax,
+                        symbolic_inner,
+                        null() as *const f64,  // Default settings
+                        null_mut() as *mut c_void  // Ignore info
+                    );
+                };
+                let symbolic_ = unsafe{$Symbolic(*symbolic_inner)};
```
mulimoen, line 101: Drop trailing underscore

suitesparse_bindings/sprs_suitesparse_umfpack/src/lib.rs
```diff
@@ -0,0 +1,380 @@
...
+//! This wrapper currently covers the double-int (DI) and double-long (DL) variations of
+//! the underlying library, while several more variations (such as for complex data type)
+//! exist in the underlying library but do not have wrappers here.
+
+use core::ptr::{null, null_mut};
+use libc::c_void;
+use sprs::{CsMatI, PermOwnedI};
+use suitesparse_umfpack_sys::*;
+
+macro_rules! umfpack_impl {
```
mulimoen, line 14: Bit of a shame all this code is in a macro. We won't get any help from clippy

suitesparse_bindings/sprs_suitesparse_umfpack/src/lib.rs
```diff
@@ -0,0 +1,380 @@
...
+
+            /// Get the number of nonzero entries in A
+            pub fn nnz(&self) -> usize {
+                self._nnz as usize
+            }
+
+            /// Get a reference to the stored matrix,
+            /// which may have had its data type converted from
+            /// what was supplied.
+            pub fn a(&self) -> &CsMatI<f64, $int> {
```
mulimoen, line 141: It is more ergonomic to return `CsMatViewI<f64, $int>` (use `self.a.view()` in the function body)

suitesparse_bindings/sprs_suitesparse_umfpack/src/lib.rs
```diff
@@ -0,0 +1,380 @@
...
+            }
+
+            /// Get raw components of the numerical factorization of `A`
+            /// * `l`: `L` matrix in CSC format
+            /// * `u`: `U` matrix in CSR format
+            /// * `p`: row permutation
+            /// * `q`: column permutation
+            /// * `rs`: inverse row scaling (divide rows of LU by these to recover PAQ)
+            /// * `dx`: unknown; this quantity is not mentioned in underlying documentation but has distinct values, so we provide it here
+            pub fn get_numeric(&self) -> (CsMatI<f64, $int>, CsMatI<f64, $int>, PermOwnedI<$int>, PermOwnedI<$int>, Vec<f64>, Vec<f64>) {
```
mulimoen, line 219: Should we return a struct from this function?

suitesparse_bindings/suitesparse_umfpack_sys/src/lib.rs
```diff
@@ -0,0 +1,28 @@
+//! Raw C bindings to some (but not all) UMFPACK functions.
+
+#[cfg(target_os = "windows")]
+pub type SuiteSparseLong = libc::c_longlong;
+#[cfg(not(target_os = "windows"))]
+pub type SuiteSparseLong = libc::c_long;
+
+pub type SuiteSparseInt = libc::c_int;
```
mulimoen, line 8: `libc::T` are in `std::ffi::T`

mulimoen (commented):
This is some great work! Just a couple of minor nits and this should be good to go.

I'll be happy to have a look at the static building later on. It should be possible to build this using e.g. `cargo check --features suitesparse_ldl_sys/static`, we can add a feature flag to more crates to make this easier to enable

suitesparse_bindings/sprs_suitesparse_umfpack/src/lib.rs
```diff
@@ -0,0 +1,380 @@
+//! Partial interface (3rd party, unaffiliated) to `SuiteSparse`'s UMFPACK solver package,
+//! covering essentials for solving problems of the form Ax=b for real A, and for
+//! recovering LU decomposition components of A for other uses.
+//!
+//! This wrapper currently covers the double-int (DI) and double-long (DL) variations of
+//! the underlying library, while several more variations (such as for complex data type)
+//! exist in the underlying library but do not have wrappers here.
+
+use core::ptr::{null, null_mut};
+use libc::c_void;
```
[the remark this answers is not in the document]
jlogan03: done, thanks!

suitesparse_bindings/sprs_suitesparse_umfpack/src/lib.rs
jlogan03: awesome! done

suitesparse_bindings/sprs_suitesparse_umfpack/src/lib.rs
jlogan03: donezo

suitesparse_bindings/sprs_suitesparse_umfpack/src/lib.rs
jlogan03: Clippy does give me commentary on the contents of the macro - might be from a recent update?

suitesparse_bindings/sprs_suitesparse_umfpack/src/lib.rs
jlogan03: Done

suitesparse_bindings/sprs_suitesparse_umfpack/src/lib.rs
jlogan03: I'm down with either option - this method allows you to mask out parts of the return which could reduce peak heap usage in cases where the compiler is able to identify that the values are never used... but for real applications you definitely need all of them except maybe dx, so it might as well be a struct

suitesparse_bindings/suitesparse_umfpack_sys/src/lib.rs
jlogan03: fixed

suitesparse_bindings/sprs_suitesparse_umfpack/src/lib.rs
jlogan03: So, tying this in with the other thought about clippy - I had to add underscores to these because clippy was giving me lints on these at the initialization of $Context. It gets tricked by repeated words that it thinks are repeated fields, like

Struct {
    x: x  // Here, it thinks the second `x` is a repeated field instead of a value with the same name
}

jlogan03:
> This is some great work! Just a couple of minor nits and this should be good to go.
> 
> I'll be happy to have a look at the static building later on. It should be possible to build this using e.g. `cargo check --features suitesparse_ldl_sys/static`, we can add a feature flag to more crates to make this easier to enable

Sounds good - I'll be interested to follow along and might give it another try myself eventually

suitesparse_bindings/sprs_suitesparse_umfpack/src/lib.rs
jlogan03: Reworked it to return a struct. Definitely cleaner this way

jlogan03:
Good to go for another round of CI - might even run to completion this time...

jlogan03:
So, I'm not able to reproduce this CAMD test failure locally (on a mac), but my best guess is that it might be a semver-non-compliant change in cc, so I rolled the patch version for suitesparse-src's cc dep back to what it was before. Other than that, I don't think anything I've changed here should have an effect.

jlogan03:
For the first failed job, that looks like a network issue. 

The second one (the failing camd tests on mac) are the same as before, and I have no idea why it's happening. I'm able to run those tests locally on mac without issue:

```
test result: ok. 28 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 6.39s

(py310) Jamess-MacBook-Pro:sprs_suitesparse_camd jlogan$ cargo test --features suitesparse_camd_sys/static -p sprs
```

jlogan03:
Ok, on further inspection, I realized that test is running on nightly - I do get that error locally running on the nightly channel, but not on stable. Seems to be a bug in nightly unrelated to the changes on this PR

suitesparse_bindings/sprs_suitesparse_umfpack/src/lib.rs
mulimoen: The tip says to have just `dx`, giving struct declarations like
```rust
                $NumericComponents {
                    l,
                    u,
                    p,
                    q,
                    rs,
                    dx
                }
```

mulimoen:
I find the random CI error strange. I reran CI from master which gives no errors: https://github.com/sparsemat/sprs/actions/runs/4705082629

I am not sure if this is a miscompilation or a bug in some code we have here.

suitesparse_bindings/sprs_suitesparse_umfpack/src/lib.rs
jlogan03: That's because I made an example to show that error - the actual definitions all have it fixed. There is an alternative to the trailing underscores, which is to go to every single location where it causes a problem and tell clippy to ignore it, but this seems cleaner

jlogan03:
> I find the random CI error strange. I reran CI from master which gives no errors: https://github.com/sparsemat/sprs/actions/runs/4705082629
> 
> I am not sure if this is a miscompilation or a bug in some code we have here.

This is running successfully on nightly on my mac now - looks like whatever was busted on nightly, they must have fixed it overnight

jlogan03:
...aaaand it's breaking again. I guess since it's a pointer alignment issue, it will probably be intermittent

jlogan03:
...and running again - definitely intermittent

jlogan03:
The benign changes to suitesparse-src that were leftover from reversing my attempt to do the static build have been removed just in case those are somehow the problem, and this is ready for another CI attempt

suitesparse_bindings/sprs_suitesparse_umfpack/src/lib.rs
jlogan03: Ok, I figured out what was going on here and was able to get rid of all the trailing underscores

jlogan03:
Figured out how to get rid of all those trailing underscores & I believe this should be ready to merge

mulimoen:
Thanks a lot for your work in this @jlogan03!

PR #329, opened 2023-05-23
Add append_outer_iter

kolstae:
We use sprs for large sparse matrices with dims > 250,000,000. Currently the options to build a matrix requires realizing a `CsVecI` or a dense vector, which is inefficient.

Adding `append_outer_iter` helps with this without adding much code.
Makes e.g. filtering out a component from a large matrix much simpler and efficient.

### DIFF
--- a/sprs/src/sparse/csmat.rs
+++ b/sprs/src/sparse/csmat.rs
@@ -563,6 +563,9 @@ impl<N, I: SpIndex, Iptr: SpIndex> CsMatI<N, I, Iptr> {
         N: Num,
         Iter: Iterator<Item = (usize, N)>,
     {
+        if let (_, Some(nnz)) = iter.size_hint() {
+            self.reserve_nnz(nnz)
+        }
         let mut nnz = self.nnz();
         let mut prev_inner_ind = None;
         for (inner_ind, val) in iter {
--- a/sprs/src/sparse/csmat.rs
+++ b/sprs/src/sparse/csmat.rs
@@ -556,7 +556,8 @@ impl<N, I: SpIndex, Iptr: SpIndex> CsMatI<N, I, Iptr> {
         self.append_outer_iter(data.iter().cloned().enumerate())
     }
 
-    /// Append an outer dim to an existing matrix, provided by an iterator
+    /// Append an outer dim to an existing matrix, increasing the size along the outer
+    /// dimension by one.
     pub fn append_outer_iter<Iter>(mut self, iter: Iter) -> Self
     where
         N: Num,
--- a/sprs/src/sparse/csmat.rs
+++ b/sprs/src/sparse/csmat.rs
@@ -722,7 +722,7 @@ impl<N, Iter: Iterator<Item = (usize, N)>> Iterator
 
         if let Some(prev_idx) = self.prev {
             assert!(
-                idx < prev_idx,
+                prev_idx < idx,
                 "index out of order. {} followed {}",
                 idx,
                 prev_idx
--- a/sprs/src/sparse/csmat.rs
+++ b/sprs/src/sparse/csmat.rs
@@ -553,35 +553,49 @@ impl<N, I: SpIndex, Iptr: SpIndex> CsMatI<N, I, Iptr> {
     where
         N: Clone + Num,
     {
-        self.append_outer_iter(data.iter().cloned().enumerate())
+        self.append_outer_iter_unchecked(
+            data.iter()
+                .cloned()
+                .enumerate()
+                .filter(|(_, val)| !val.is_zero()),
+        )
     }
 
     /// Append an outer dim to an existing matrix, increasing the size along the outer
     /// dimension by one.
-    pub fn append_outer_iter<Iter>(mut self, iter: Iter) -> Self
+    ///
+    /// **Note** Panics if the iterator is **not** strictly ordered by ascending index
+    pub fn append_outer_iter<Iter>(self, iter: Iter) -> Self
     where
         N: Num,
+        Iter: Iterator<Item = (usize, N)>,
+    {
+        self.append_outer_iter_unchecked(AssertOrderedIterator {
+            prev: None,
+            iter: iter.filter(|(_, val)| !val.is_zero()),
+        })
+    }
+
+    /// Append an outer dim to an existing matrix, increasing the size along the outer
+    /// dimension by one.
+    ///
+    /// **Note** The iterator must be strictly ordered by ascending index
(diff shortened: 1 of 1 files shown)


jlogan03:
Thanks for the writeup! I'm traveling for a couple of days, but I'll take a look and run some quick bench comparisons later in the week, if Magnus doesn't get to it first. 

Looks good from a quick skim - I just want to check that the additional branch point inside the loop (which I agree is necessary since the provided iterator isn't guaranteed to be sorted) doesn't scare away any optimizations. Given that there is already a branch point in the loop, it's unlikely that there would be any change.

sprs/src/sparse/csmat.rs
```diff
@@ -549,18 +549,38 @@ impl<N, I: SpIndex, Iptr: SpIndex> CsMatI<N, I, Iptr> {
...
+        Iter: Iterator<Item = (usize, N)>,
     {
         let mut nnz = self.nnz();
-        for (inner_ind, val) in data.iter().enumerate() {
-            if *val != N::zero() {
+        let mut prev_inner_ind = None;
+        for (inner_ind, val) in iter {
+            if val != N::zero() {
+                if let Some(p_ind) = prev_inner_ind {
+                    assert!(p_ind < inner_ind, "inner index order");
```
jlogan03, line 570:
```suggestion
                    assert!(p_ind < inner_ind, "Indices in provided data are out of order. To fix this, sort by index in ascending order.");
```

sprs/src/sparse/csmat.rs
```diff
@@ -549,18 +549,38 @@ impl<N, I: SpIndex, Iptr: SpIndex> CsMatI<N, I, Iptr> {
...
                 self.indices.push(I::from_usize(inner_ind));
-                self.data.push(val.clone());
+                self.data.push(val);
                 nnz += 1;
             }
         }
+        if let Some(last_inner_ind) = prev_inner_ind {
+            assert!(
+                last_inner_ind < self.inner_dims(),
+                "inner index out of range"
```
jlogan03, line 581: Similarly, to aid troubleshooting, maybe something like this:

```suggestion
                "Indices in provided data exceed size of inner dimension. To fix this, provide data that fits within the existing inner dimension."
```

sprs/src/sparse/csmat.rs
kolstae: I'm personally not a big fan of elaborate error-messages. Perhaps provide some data to help out instead? What about something like
```rust
      assert!(p_ind < inner_ind, "Indices in provided data are out of order. Got {} then {}", p_ind, inner_ind);
```

kolstae:
> Looks good from a quick skim - I just want to check that the additional branch point inside the loop (which I agree is necessary since the provided iterator isn't guaranteed to be sorted) doesn't scare away any optimizations. Given that there is already a branch point in the loop, it's unlikely that there would be any change.

Could we make a trait signaling and/or verifying it's already ordered? This could be implemented as a noop for `Enumerated`.

sprs/src/sparse/csmat.rs
```diff
@@ -549,18 +549,38 @@ impl<N, I: SpIndex, Iptr: SpIndex> CsMatI<N, I, Iptr> {
...
     /// Append an outer dim to an existing matrix, compressing it in the process
-    pub fn append_outer(mut self, data: &[N]) -> Self
+    pub fn append_outer(self, data: &[N]) -> Self
     where
         N: Clone + Num,
+    {
+        self.append_outer_iter(data.iter().cloned().enumerate())
+    }
+
+    /// Append an outer dim to an existing matrix, provided by an iterator
```
mulimoen, line 559:
```suggestion
    /// Append an outer dim to an existing matrix, increasing the size along the outer dimension by one.
    ///
    /// This version is the more general version of `Self::append_outer` and `Self::append_outer_csvec`
    /// The iterator supplied should yield the inner index and the data element to be added.
    /// 
    /// # Panics
    /// * If indices are not in monotonically increasing order
    /// * If indices are out of bounds
```

mulimoen (commented):
Looks great as an abstraction of the more specialised appends. Maybe `append_outer_csvec` could call the iterator version too?

sprs/src/sparse/csmat.rs
```diff
@@ -549,18 +549,42 @@ impl<N, I: SpIndex, Iptr: SpIndex> CsMatI<N, I, Iptr> {
...
+
+    /// Append an outer dim to an existing matrix, increasing the size along the outer
+    /// dimension by one.
+    pub fn append_outer_iter<Iter>(mut self, iter: Iter) -> Self
+    where
+        N: Num,
+        Iter: Iterator<Item = (usize, N)>,
+    {
+        let (lower, opt_upper) = iter.size_hint();
+        self.reserve_nnz(opt_upper.unwrap_or(lower));
```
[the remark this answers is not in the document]
mulimoen: Was thinking something like
```rust
if let (_, Some(nnz)) = iter.size_hint() {
   self.reserve_nnz(nnz)
}
```

kolstae:
How about this approach? Move the validation of order into an explicit struct.

The `AssertOrderedIterator` should probably be moved somewhere else, but stuck it there for convenience for now.

mulimoen:
@kolstae I added a commit directly, does the `unsafe` keyword pose any difficulties for you?

kolstae:
> @kolstae I added a commit directly, does the `unsafe` keyword pose any difficulties for you?

No, not at all. `unsafe`'s a better way so make things clear.

sprs/src/sparse/csmat.rs
```diff
@@ -549,17 +549,69 @@ impl<N, I: SpIndex, Iptr: SpIndex> CsMatI<N, I, Iptr> {
     }
 
     /// Append an outer dim to an existing matrix, compressing it in the process
-    pub fn append_outer(mut self, data: &[N]) -> Self
+    pub fn append_outer(self, data: &[N]) -> Self
     where
-        N: Clone + Num,
+        N: Clone + Zero,
```
kolstae, line 554: Will this (`Num` -> `Zero`) change break existing code though?

mulimoen: As `Zero` is a subtrait I would say no https://docs.rs/num-traits/latest/num_traits/trait.Num.html

mulimoen:
Thanks for adding this @kolstae!

mulimoen:
A new version of `sprs` (0.11.1) is released containing this PR

PR #331, opened 2023-08-13
Include license files in published crates

ankane:
Hey, thanks for this library!

This PR symlinks the license files so they're included in the published crates (like with sprs 0.11.0).

### DIFF
--- a/sprs-rand/LICENSE-APACHE
+++ b/sprs-rand/LICENSE-APACHE
@@ -1,0 +1,1 @@
+../LICENSE-APACHE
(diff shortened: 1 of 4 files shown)


mulimoen:
Thanks for the PR. I would think the licenses are fairly well known and not necessary to include. Is there a benefit in adding these files to the published blobs?

ankane:
Hi @mulimoen, both licenses mention including it with copies of the software, and it's nice for things like `cargo vendor` (https://github.com/microsoft/windows-rs/issues/1587#issuecomment-1063010186).

mulimoen:
I see. Thanks for the explanation and the PR!

ankane:
Great, thanks!

PR #334, opened 2023-09-04
Add more permutation functionality

jlogan03:
* Add private functions `permute_inner` and `permute_outer` that separate the permutation of inner and outer dimensions as found in `transform_mat_papt`
* Add public functions `permute_rows` and `permute_cols` that select either `permute_inner` or `permute_outer` depending on storage
* Add public function `transform_mat_paq` that mirrors `transform_mat_papt` but allows different permutations for rows and cols and allows nonsquare matrices
* Use `transform_mat_paq` to test umfpack bindings' numeric factorization outputs
  * Incidentally macroize umfpack tests to reduce boilerplate overhead

### DIFF
--- a/sprs/src/lib.rs
+++ b/sprs/src/lib.rs
@@ -109,8 +109,8 @@ pub use crate::mul_acc::MulAcc;
 pub use crate::sparse::symmetric::is_symmetric;
 
 pub use crate::sparse::permutation::{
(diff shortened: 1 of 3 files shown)


mulimoen (approved):
I can't see anything obviously wrong so I'll approve this

Issue #339, opened 2023-09-30
Linear solver interfaces


jlogan03:
I have a rough working prototype of serial, sparse-sparse BiCGSTAB [here](https://github.com/jlogan03/sprs/blob/jlogan/bicgstab/sprs/src/sparse/linalg/bicgstab.rs) and it's got me thinking about how to structure interfaces to solvers.

It seems like there are a few specific actions that users need regularly that might be supported in different ways by different kinds of solvers -
* Just solve a system with minimal fiddling
* Prep a solver, keeping factorization, preconditioners, etc. to use for solving a system multiple times with different right-hand side
* Advance an iterative solver, then do something with the latest error estimate and solution
    * Convergence plots, checkpointing, etc
* Restart an iterative solver's internal state, preserving the latest solution
* Swap components of an initialized solver with minimum re-computation
    * Like keeping a symbolic factorization and using it to swap in a new `A` matrix for FEM with deformation
    * ...or swapping a new `A` or rhs into a partially-converged iterative solver to capture a nonlinear update

To that end, I'm thinking each solver should be associated with both 
* A standalone function that provides the fuss-free solve, and
* A struct that keeps a workspace and implements methods for manipulating that workspace

All pretty vanilla so far.

Now, there's a pretty different pattern of usage between direct and iterative solvers, and I can't think of a way to stuff them into the same trait. So with that, we'd have two traits, say, DirectSolver and IterativeSolver, guiding the implementation of the solver structs, and possibly another two guiding the implementation of the convenience functions.

This leaves us with
* Two traits guiding implementation of direct and iterative solvers
* Optionally, another two traits or type aliases guiding the convenience functions
* For each solver,
    * A workspace struct implementing either DirectSolver or IterativeSolver
    * A convenience function, which either
        * is part of the impl on the struct
        * implements a trait or matches a provided type alias
        * is freeform and unconstrained in its construction

@mulimoen - I'm curious what your thoughts are on
* Whether now is a good time to implement such a structure, or whether new solvers should be added independently for now
* Whether there are important use-cases that I haven't captured
* Whether the convenience functions should be standalone (and maybe have their own traits or type aliases) or whether they should be part of the impl on the solver workspace structs
* Any other thoughts about the proposed structure

mulimoen:
The implementation of BiCGSTAB would be great to have in this crate.

[...]

PR #340, opened 2023-10-08
Implement BiCGSTAB

jlogan03:
Some thoughts about this implementation -

Pros
* It runs and is even fairly correct and a bit documented
* Implemented on f32 and f64 with generic indices

Cons
* Doesn't represent any particular reusable abstract pattern of iterative solver interface, and only makes a little progress toward the goals of #339 
* Solver call signature requires the turbof***
  * I haven't found any way around this - even when it's obvious from the inputs and outputs which variant should be selected, it still fails to select the correct variant without explicit type annotation
* Macros make things worse for the users in terms of readability, understanding errors, etc
* Not fully generic on value type
* Not generic on sparse vs. dense RHS and initial guess
* Can't be implemented directly on Complex<T> types
* Allocates then drops several vectors on each iteration which have the same sparsity at every iteration and could reuse the same memory if we had a mechanism for this
  * In practice, this isn't as bad as it sounds since the first few iterations will clear out a correctly sized region of heap s.t. the subsequent allocations are cheap to perform... as long as there is not much else happening on the machine

I spent a solid 8 hours attempting to make a version that is generic over the value type. This did not go well. 

Some of the implementations of the underlying matrix-vector and vector-scalar operations are not fully generic, so the trait bounds on the impl end up being more than 10 lines long and incredibly unclear as to which types will and will not be compatible, and the type system ultimately fails on overflow due to inferring generic outputs across iterations in the solver loop. 

This could all be fixed to some extent, but would require quite a bit of work to go back into the core of the crate and make a single definition of what trait bounds constitute a mathematical field over which we can do linear algebra, then propagate that definition to every struct and impl in the crate, which I believe is out of scope for this PR.

### DIFF
--- a/sprs/src/sparse/linalg/bicgstab.rs
+++ b/sprs/src/sparse/linalg/bicgstab.rs
@@ -1,0 +1,391 @@
+//! Stabilized bi-conjugate gradient solver for solving Ax = b with x unknown. Suitable for non-symmetric matrices.
+//! A simple, sparse-sparse, serial, un-preconditioned implementation.
+//!
+//! # References
+//! The original paper, which is thoroughly paywalled but widely referenced:
+//!
+//! ```text
+//! H. A. van der Vorst,
+//! “Bi-CGSTAB: A Fast and Smoothly Converging Variant of Bi-CG for the Solution of Nonsymmetric Linear Systems,”
+//! SIAM Journal on Scientific and Statistical Computing, Jul. 2006, doi: 10.1137/0913035.
+//! ```
+//!
+//! A useful discussion of computational cost and convergence characteristics for the CG
+//! family of algorithms can be found in the paper that introduces QMRCGSTAB, in Table 1:
+//!
+//! ```text
+//! T. F. Chan, E. Gallopoulos, V. Simoncini, T. Szeto, and C. H. Tong,
+//! “A Quasi-Minimal Residual Variant of the Bi-CGSTAB Algorithm for Nonsymmetric Systems,”
+//! SIAM J. Sci. Comput., vol. 15, no. 2, pp. 338–347, Mar. 1994, doi: 10.1137/0915023.
+//! ```
+//!
+//! A less-paywalled pseudocode variant for this solver (as well as CG aand CGS) can be found at:
+//! ```text
+//! https://utminers.utep.edu/xzeng/2017spring_math5330/MATH_5330_Computational_Methods_of_Linear_Algebra_files/ln07.pdf
+//! ```
+//!
+//! # Example
+//! ```rust
+//! use sprs::{CsMatI, CsVecI};
+//! use sprs::linalg::bicgstab::BiCGSTAB;
+//!
+//! let a = CsMatI::new_csc(
+//!     (4, 4),
+//!     vec![0, 2, 4, 6, 8],
+//!     vec![0, 3, 1, 2, 1, 2, 0, 3],
+//!     vec![1.0, 2., 21., 6., 6., 2., 2., 8.],
+//! );
+//!
+//! // Solve Ax=b
+//! let tol = 1e-60;
+//! let max_iter = 50;
+//! let b = CsVecI::new(4, vec![0, 1, 2, 3], vec![1.0; 4]);
+//! let x0 = CsVecI::new(4, vec![0, 1, 2, 3], vec![1.0, 1.0, 1.0, 1.0]);
+//!
+//! let res = BiCGSTAB::<'_, f64, _, _>::solve(
+//!     a.view(),
+//!     x0.view(),
+//!     b.view(),
+//!     tol,
+//!     max_iter,
+//! )
+//! .unwrap();
+//! let b_recovered = &a * &res.x();
+//!
+//! println!("Iteration count {:?}", res.iteration_count());
+//! println!("Soft restart count {:?}", res.soft_restart_count());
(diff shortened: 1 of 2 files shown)


sprs/src/sparse/linalg/bicgstab.rs
```diff
@@ -0,0 +1,391 @@
...
+    rhat: CsVecI<T, I>, // Arbitrary w/ dot(rhat, r) != 0
+    p: CsVecI<T, I>,
+    // Intermediate scalars
+    rho: T,
+}
+
+macro_rules! bicgstab_impl {
+    ($T: ty) => {
+        impl<'a, I: SpIndex, Iptr: SpIndex> BiCGSTAB<'a, $T, I, Iptr> {
+            /// Initialize the solver with a fresh error estimat
```
mulimoen, line 117:
```suggestion
            /// Initialize the solver with a fresh error estimate
```

sprs/src/sparse/linalg/bicgstab.rs
```diff
@@ -0,0 +1,391 @@
...
+        println!("Hard restart count {:?}", res.hard_restart_count());
+
+        // Make sure the solved values match expectation
+        for (input, output) in
+            b.to_dense().iter().zip(b_recovered.to_dense().iter())
+        {
+            assert!(
+                (1.0 - input / output).abs() < tol,
+                "Solved output did not match input"
+            );
```
mulimoen, line 348:
```suggestion
            assert!(
                approx::relative_eq!(input, output, max_relative = tol),
                "Solved output did not match input"
            );

```

mulimoen: And `cargo add --dev approx` to add it to `Cargo.toml`

sprs/src/sparse/linalg/bicgstab.rs
```diff
@@ -0,0 +1,391 @@
...
+        println!("Iteration count {:?}", res.i());
+        println!("Soft restart count {:?}", res.soft_restart_count());
+        println!("Hard restart count {:?}", res.hard_restart_count());
+
+        // Make sure the solved values match expectation
+        for (input, output) in
+            b.to_dense().iter().zip(b_recovered.to_dense().iter())
+        {
+            assert!(
+                (1.0 - input / output).abs() < tol,
```
mulimoen, line 386:
```suggestion
                approx::relative_eq!(input, output, max_relative = tol),
```

sprs/src/sparse/linalg/bicgstab.rs
```diff
@@ -0,0 +1,391 @@
...
+            }
+
+            /// Set the minimum value of `rho` to trigger a soft restart
+            pub fn with_restart_threshold(mut self, thresh: $T) -> Self {
+                self.soft_restart_threshold = thresh;
+                self
+            }
+
+            /// Iteration number
+            pub fn i(&self) -> usize {
```
mulimoen, line 238: Should this be given a more descriptive name such as `iteration_number`?

mulimoen: But I have not read the papers and I don't know if this is standard lingo

sprs/src/sparse/linalg/bicgstab.rs
```diff
@@ -0,0 +1,391 @@
+//! Stabilized bi-conjugate gradient solver. Suitable for non-symmetric matrices.
```
mulimoen, line 1:
```suggestion
//! Stabilized bi-conjugate gradient solver for solving Ax = b with x unknown. Suitable for non-symmetric matrices.
```

sprs/src/sparse/linalg/bicgstab.rs
```diff
@@ -0,0 +1,391 @@
...
+//! println!("Iteration count {:?}", res.i());
+//! println!("Soft restart count {:?}", res.soft_restart_count());
+//! println!("Hard restart count {:?}", res.hard_restart_count());
+//!
+//! // Make sure the solved values match expectation
+//! for (input, output) in
+//!     b.to_dense().iter().zip(b_recovered.to_dense().iter())
+//! {
+//!     assert!(
+//!         (1.0 - input / output).abs() < tol,
```
mulimoen, line 64:
```suggestion
//!         approx::relative_eq!(input, output, max_relative = tol
```

mulimoen (approved):
This is some fantastic work and it's great to see a proper general solver. This will be immensely useful to users.

Making everything generic has many drawbacks, the ergonomics can get much worse and it takes a lot longer to write for dubious benefit, the macro is great for implementing the two most useful impls and we can deal with `Complex` if anybody asks for it.

I only have some minor nits on this PR, I consider it good for a merge in the current state

sprs/src/sparse/linalg/bicgstab.rs
jlogan03: It's not standard lingo, just me being lazy - updated it to `iteration_count` for consistency with the other `*_count` values stored by the solver

jlogan03:
@mulimoen I like the approx usage for testing, but had to roll it back because we don't keep approx as a default feature so it crashes the tests. I'd be happy to bring it back if we want to update either default features to include approx, or to include the approx feature for testing

mulimoen:
I don't feel strongly about approx, bit of a pain to enable it for doctests. A bit of resolver history: The resolver previously merged features across build types, so having approx on during tests would enable it also for regular builds. Since we are using resolver v2 we can now use it for tests only

PR #369, opened 2024-10-22
maximum allowed version for sprs incremented to 0.16.1

HenrikJStromberg:


sprs-ldl/Cargo.toml
```diff
@@ -18,4 +18,4 @@ sprs_suitesparse_ldl = { version = "0.8.0", path = "../suitesparse_bindings/sprs
 sprs_suitesparse_camd = { version = "0.3.0", path = "../suitesparse_bindings/sprs_suitesparse_camd", optional = true }
 
 [dev-dependencies]
-ndarray = ">=0.15.0, <=0.16"
+ndarray = ">=0.15.0, <=0.16.1"
```
jofas, line 21: I think kpreid's [suggestion on URLO](https://users.rust-lang.org/t/cargo-selects-conficting-ndarray-versions/120108/4?u=jofas) should be applied here, namely changing this to

```suggestion
ndarray = ">=0.15.0, <0.17"
```

to support all `0.16.*` versions, not just `0.16.0` and `0.16.1`.

Although, if it is not necessarily considered a feature to support both `0.15` and `0.16` versions, I would probably just bump this to `^0.16.0` to avoid any maintenance overhead and potential pitfalls when making changes to these packages:

```suggestion
ndarray = "0.16"
```

sprs/Cargo.toml
```diff
@@ -21,7 +21,7 @@ multi_thread = ["dep:rayon", "dep:num_cpus"]
 
 [dependencies]
 num-traits = "0.2.0"
-ndarray = ">=0.15, <=0.16"
+ndarray = ">=0.15, <=0.16.1"
```
jofas, line 24: Same suggestions as in comment above:

```suggestion
ndarray = ">=0.15.0, <0.17"
```

```suggestion
ndarray = "0.16"
```

mulimoen:
Thanks @jofas!

PR #375, opened 2025-01-20
Fix Index conversion turning OOB

### LINKED ISSUE #374: Bug in sparse-sparse multiplication and/or transposition with small index type
### Short description

I am using a `CsMatI<f32, I = u16, Indptr = usize>` (CSR) for saving memory when constructing a matrix with approx. ~100 million rows and ~50 columns. When multiplying the matrix with its own transpose, I am getting too many zeroes – the matrix is zero in places it definitely shouldn't be. I receive no compile-time error, and no runtime error, the library just silently produces an incorrect error.

Since building/indexing the matrix works perfectly in itself as long as I do not involve the transpose, I suspect the root cause will probably have something to do with the transposition and the column index type not being large enough to store row indexes/index pointers. If this is the case, however, it should probably result in preferably a compile-time error, and definitely at least a run-time error.

### Reproducer

Here is a minimal reproducer that constructs a matrix, multiplies it with its own transpose using both `sprs` and by converting it (and its transpose) to a dense `ndarray::Array2`, and comparing the number of zeros in the results. This behavior is exhibited by the latest (as of now) version of `sprs` (0.11.2) and `ndarray` (0.16.1), but I originally encountered it in `sprs` 0.11.1 with `ndarray` 0.15.6.

```rust
use sprs::CsMatI;
use ndarray::Array2;

type Mat = CsMatI<f32, u16, usize>;

fn dense_nnz(arr: &Array2<f32>) -> usize {
    arr.iter().map(|&f| usize::from(f != 0.0)).sum()
}

fn main() {
    // create matrix so that row index doesn't fit into column index type after transposition
    let mut x_sprs = Mat::zero((1_usize << 18, usize::from(1_u16 << 4)));
    x_sprs.insert(1_usize << 17, usize::from(1_u16 << 2), 1.0);

    // compute X^T * X in the sparse representation
    let x_t_sprs = x_sprs.transpose_view();
    let x_t_x_sprs = &x_t_sprs * &x_sprs;

    // compute X^T * X in the dense representation
    let x_dense = x_sprs.to_dense();
    let x_t_x_dense = x_dense.t().dot(&x_dense);

    // convert sparse result to dense representation for comparison
    let x_t_x_sprs_to_dense = x_t_x_sprs.to_dense();

    println!("Sparse:          NNZ(X)       = {}", x_sprs.nnz());
    println!("Sparse:          NNZ(X^t * X) = {}", x_t_x_sprs.nnz());
    println!("Dense:           NNZ(X)       = {}", dense_nnz(&x_dense));
    println!("Dense:           NNZ(X^t * X) = {}", dense_nnz(&x_t_x_dense));
    println!("Sparse-to-dense: NNZ(X^t * X) = {}", dense_nnz(&x_t_x_sprs_to_dense));
}
```

### 
Actual:
```
Sparse:          NNZ(X)       = 1
Sparse:          NNZ(X^t * X) = 0
Dense:           NNZ(X)       = 1
Dense:           NNZ(X^t * X) = 1
Sparse-to-dense: NNZ(X^t * X) = 0
```
Expected:
```
Sparse:          NNZ(X)       = 1
Sparse:          NNZ(X^t * X) = 1
Dense:           NNZ(X)       = 1
Dense:           NNZ(X^t * X) = 1
Sparse-to-dense: NNZ(X^t * X) = 1
```

mulimoen:
Fix #374

### DIFF
--- a/sprs/src/sparse/csmat.rs
+++ b/sprs/src/sparse/csmat.rs
@@ -1788,6 +1788,11 @@ pub mod raw {
 
         assert!(indptr.iter().all(num_traits::Zero::is_zero));
 
+        assert!(
+            I::try_from_usize(mat.rows()).is_some(),
+            "Index type is not large enough to hold the number of rows requested (I::max_value={:?} vs. required {})", I::max_value(), mat.rows(),
+        );
+
         for vec in mat.outer_iterator() {
             for (inner_dim, _) in vec.iter() {
                 indptr[inner_dim] += Iptr::one();
@@ -1805,10 +1810,11 @@ pub mod raw {
         }
 
         for (outer_dim, vec) in mat.outer_iterator().enumerate() {
+            let outer_dim = I::from_usize_unchecked(outer_dim);
             for (inner_dim, val) in vec.iter() {
                 let dest = indptr[inner_dim].index();
                 data[dest] = val.clone();
-                indices[dest] = I::from_usize_unchecked(outer_dim);
+                indices[dest] = outer_dim;
                 indptr[inner_dim] += Iptr::one();
             }
         }
--- a/sprs/tests/gh374.rs
+++ b/sprs/tests/gh374.rs
@@ -1,0 +1,33 @@
+use ndarray::Array2;
+use sprs::CsMatI;
+
+type Mat = CsMatI<f32, u16, usize>;
+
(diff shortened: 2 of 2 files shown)


H2CO3:
Great, thanks!

